PR #32, opened 2026-07-15
fix(ssh): lost keystrokes inside sudo -i with shell integration enabled

ezhkov-ph:
## Problem

With shell integration enabled (the default), typing inside `sudo -i` on hosts where sudo runs in `use_pty` mode (default on modern Ubuntu/Debian) loses most keystrokes: a key has to be pressed several times for one character to register, Backspace behaves the same way, and pasting arrives garbled (literal `200~`/`201~` fragments). Outside `sudo` the same session types perfectly, and other SSH clients are unaffected on the same host.

## Root cause

The SSH/WSL shell-integration wrappers re-attached the final shell's stdin by reopening the terminal **by path** (`</dev/tty`). That yields a *different file description* of the tty. sudo's `use_pty` relay then fails to recognize the shell's stdin as the user's terminal, input handling splits between sudo's pty relay and the command, and each keystroke races between two readers — most lose.

Confirmed bidirectionally on Ubuntu:
- in a broken session, `exec bash -i <&2` fixes typing under `sudo -i`;
- in a healthy session, `exec bash -i </dev/tty` reproduces the loss.

Diagnostic logging also confirmed the client sends every byte exactly once (sudo password entry — canonical mode — was always lossless; loss started exactly when the raw-mode pty relay took over).

## Fix

Duplicate the inherited stderr descriptor (`<&2`) instead — stderr still holds the pty file description sshd created. This matches what the persistent (tmux/screen) wrapper already does, where reopening by path was independently documented as breaking modern tmux. Applied to all `SSH_WRAPPER` and `WSL_WRAPPER` branches (bash/zsh/fish/sh).

## Also included

`feat(terminal)`: an off-by-default **Plain paste** toggle (xterm.js `ignoreBracketedPasteMode`) for remote endpoints that advertise bracketed paste but mishandle it — serial consoles, legacy network gear, minimal busybox shells. No behavior change unless enabled.

## Testing

- `cargo test shell_integration` — 14 passed
- `cargo fmt --all --check` — clean
- `vitest run` — 317 passed (87 files)
- Manually verified on Ubuntu over SSH: `sudo -i` typing/Backspace/paste now correct with shell integration and persistent sessions enabled; plain `request_shell` sessions unchanged

### DIFF
--- a/src-tauri/src/shell_integration.rs
+++ b/src-tauri/src/shell_integration.rs
@@ -170,17 +170,26 @@ function global:prompt {\n\
 /// safe to send regardless of the remote login shell's syntax (fish/csh
 /// would otherwise choke on POSIX case/heredoc).
 ///
-/// The `</dev/tty` on every exec is load-bearing: when run via
+/// The `<&2` on every exec is load-bearing: when run via
 /// `echo b64 | base64 -d | sh`, the inner sh's stdin is the pipe from
-/// base64. After exec, the new shell inherits that already-closed pipe and
-/// would immediately exit on EOF (printing "exit" and looping reconnect).
-/// Reopening stdin from /dev/tty restores the real PTY.
+/// base64. After exec, the new shell would inherit that already-closed pipe
+/// and immediately exit on EOF (printing "exit" and looping reconnect).
+/// Duplicating stderr — which still holds the pty file description sshd
+/// created — restores the real PTY.
+///
+/// Do NOT reopen by path (`</dev/tty`): that creates a *different* file
+/// description of the tty, and sudo's `use_pty` mode (default on modern
+/// Ubuntu/Debian) then fails to recognize the shell's stdin as the user's
+/// terminal. Input handling splits between sudo's pty relay and the command,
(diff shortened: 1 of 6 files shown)


kipavy:
Merged into `dev`. Excellent debugging on this one — the `<&2` vs `</dev/tty` file-description distinction and the bidirectional bisection made the root cause unambiguous. It also lines the SSH/WSL wrappers up with what `persistent_exec_command` has always done, so the two paths are finally consistent.

Two small follow-ups added on top before merging:
- **`ru` locale keys** for the Plain Paste setting (title/desc + toggle label) — the Russian locale (#26) is on `dev` now, so the strings could go in.
- **A regression test** (`wrappers_reattach_pty_via_stderr_not_dev_tty`) asserting both wrappers re-attach via `<&2` and never reopen by path — so a future edit can't silently reintroduce the keystroke loss. The core fix wasn't covered otherwise.

Verified on the merged tree: cargo fmt clean, `cargo test shell_integration` 15/15, tsc clean, full vitest (318) green.

Closing since the commits are on `dev` and will reach `main` at the next release. Thanks again — both the fix and the write-up were top-notch. 🙏

Issue #50, opened 2026-07-24
When editing a file, is it not possible to use the touchpad to scroll?


zwinkle:
I tried running the `nano` command on a file in a VM I’m accessing remotely via SSH, but I found the user experience to be less than ideal when I tried scrolling with the touchpad—the cursor wouldn’t move, so I had to use the arrow keys on the keyboard. Is this intended behavior, or can I request this feature?

Oh, and one more thing. When I select some text, it always copies it, even though I’ve disabled “Select to copy.” It’s not working as it should. I want to be able to select text, and for the selection to disappear very quickly—the current behavior is quite annoying to me.

<img width="925" height="610" alt="Image" src="https://github.com/user-attachments/assets/770e9993-e1ca-46de-b2a7-763c88fc1ebc" />

kipavy:
Hi @zwinkle — thanks for the detailed report (and the screenshot)! Both issues are now fixed and merged:

**1. Scrolling in full-screen apps (`nano`, `less`, `vim`, …) with the touchpad/wheel.**
These apps run in the terminal's "alternate screen", which has no scrollback of its own, so the wheel had nothing to move — which is why the cursor/view wouldn't budge. Voltius now translates wheel/touchpad scrolling into the app's own up/down navigation there, with proper accumulation so even small, gentle touchpad movements scroll smoothly. Apps that handle the mouse themselves (e.g. `vim` with `set mouse=a`) keep working as before, and normal scrollback is unchanged.

**2. "Select to Copy" not being respected.**
There was a bug where selecting text always copied to the clipboard even with the setting turned off. That's fixed — with "Select to Copy" off, you can now select text freely without it being copied. (Explicit copy via `Ctrl+Shift+C` still works regardless.)

Both fixes are merged and will go out in the **next release, likely within the coming days**. I'll follow up here once it's shipped. Thanks again for flagging these! 🙏

kipavy:
Fixed in v0.13.0 — see the [release notes](https://github.com/VoltiusApp/voltius/releases/tag/v0.13.0).

PR #53, opened 2026-07-24
fix(terminal): scroll alt-screen apps with wheel/touchpad; respect Select-to-Copy (#50)

kipavy:
Fixes the two terminal issues reported in #50.

## 1. Wheel / touchpad scroll does nothing in full-screen apps (nano/less/vim)

These programs run in xterm's **alternate screen buffer**, where there's no scrollback to move. xterm 6.0 has no built-in alternate-scroll option, and its internal wheel path dampens sub-50px touchpad deltas toward zero — so touchpad scrolling in `nano` was effectively a no-op (the exact symptom in the issue).

**Fix:** a custom wheel handler (`useTerminal.ts`) that, when the alt buffer is active and the app isn't tracking the mouse, translates the wheel delta into `Up`/`Down` arrow presses:

- **Sub-row accumulation** — touchpads emit many tiny pixel deltas; a carried remainder means slow scrolling still advances a line once enough pixels accumulate.
- **Fling clamp** — a single event moves at most one page, so a hard flick can't flood the PTY.
- **Stays out of the way** — normal buffer keeps native scrollback (returns `true`); apps that track the mouse (e.g. vim `set mouse=a`) keep getting real mouse events; `applicationCursorKeysMode` is honored so DECCKM apps get `ESC O A/B`.
- **`preventDefault`** so the gesture doesn't also scroll an ancestor.
- Synthesized arrows route through the **same broadcast path** as typed input, so split-pane broadcast behaves identically for scroll and keystrokes.

The row math is a pure, unit-tested core (`terminalWheelCore.ts`) covering pixel/line/page modes, carry accumulation, direction reversal, the fling clamp, and degenerate inputs.

## 2. Selecting text always copies, even with "Select to Copy" off

`handleMouseUp` in `terminalClipboard.ts` wrote the selection to the clipboard unconditionally — only the feedback badge checked the toggle (default `true`). So a user who turned the setting **off** still got copy-on-select.

**Fix:** gate the clipboard write on `getToggle("select-to-copy")`. Explicit `Ctrl+Shift+C` still copies regardless.

> Note: the issue also mentions wanting the selection to "disappear very quickly." That reads as ambiguous, so this PR addresses the concrete, unambiguous bug (toggle ignored) and leaves selection-clearing behavior unchanged — happy to follow up if a specific behavior is wanted.

## Testing
- New `terminalWheelCore.test.ts` (pure core).
- Full suite green: **853 passed**; `tsc --noEmit` clean.

### DIFF
--- a/src-tauri/src/commands/sync.rs
+++ b/src-tauri/src/commands/sync.rs
@@ -208,7 +208,6 @@ pub fn backup_export(
         }
     }
     let data = state.export_all()?;
-    let mut files = files;
     let mut secrets = data.secrets;
     let mut clocks = data.clocks;
     let excluded: HashSet<String> = excluded_ids.unwrap_or_default().into_iter().collect();
--- a/src/components/terminal/terminalClipboard.ts
+++ b/src/components/terminal/terminalClipboard.ts
@@ -88,6 +88,10 @@ export function attachTerminalClipboard(
   };
 
   const handleMouseUp = (e: MouseEvent) => {
+    // Copy-on-select is opt-out: when the "Select to Copy" toggle is off, a
+    // selection must NOT touch the clipboard (#50). Explicit Ctrl+Shift+C still
+    // copies regardless.
+    if (!getToggle("select-to-copy")) return;
     setTimeout(() => {
       const sel = term.getSelection();
       if (sel) {
--- a/src/components/terminal/terminalWheelCore.test.ts
+++ b/src/components/terminal/terminalWheelCore.test.ts
@@ -1,0 +1,41 @@
+import { wheelToRows } from "./terminalWheelCore.ts";
+import { test } from "vitest";
+
+test("terminalWheelCore", async () => {
+function assertEqual<T>(actual: T, expected: T, msg: string): void {
+  if (JSON.stringify(actual) !== JSON.stringify(expected)) { console.error(`FAIL ${msg}: expected ${JSON.stringify(expected)}, got ${JSON.stringify(actual)}`); throw new Error(msg); }
+}
+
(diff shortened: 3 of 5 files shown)


kipavy:
## ✅ Manual e2e validation (headless dev build, tauri-driver)

Drove the live debug app and exercised both fixes through real DOM/pointer input in a `less` (alternate-screen) + local shell session. Both pass.

### 1. Alt-screen wheel / touchpad scroll
Opened `less -N /tmp/f.txt` (500 lines) — alternate buffer, the exact issue-#50 scenario (`nano` behaves identically).

| Input | Result |
|---|---|
| Wheel down `deltaY 160` | top line **1 → 11** (10 rows) |
| **10 × tiny `deltaY 6`** (each < xterm's 50px dampening floor) | top **11 → 14** (accumulated 3 rows) — *the precise touchpad bug; stock xterm floors these to zero* |
| Wheel up `deltaY -112` | scrolled back up 6 rows |

`defaultPrevented: true` on every event confirms the new `attachCustomWheelEventHandler` is the authoritative path.

### 2. Select-to-Copy respects the toggle
Set a clipboard sentinel, then drag-selected a line of terminal output.

| "Select to Copy" | Selection made? | Clipboard | Copy badge |
|---|---|---|---|
| **On** (default) | yes | overwritten with selected text | shown ✓ |
| **Off** | yes (text highlighted) | **unchanged** (sentinel intact) | not shown ✓ |

So with the toggle off the selection is still made but nothing is copied — the reported bug is fixed, and copy-on-select still works when the toggle is on. `Ctrl+Shift+C` remains a separate explicit-copy path.

_Validation only — no code changed since the reviewed commit._

PR #78, opened 2026-07-29
feat(i18n): add Simplified Chinese (zh) translation

CoconutHR:
## Summary
- Adds Simplified Chinese (zh) locale support
- 23 new locale files under
- Registers zh in  and 
- All i18n tests pass (key parity verified against English)
- TypeScript compiles with zero errors

## Checklist
- [x] All 23 English locale files have matching zh translations
- [x]  passes (no orphaned/missing keys)
- [x]  passes
- [x]  passes


kipavy:
Thanks for this — the mechanical quality is genuinely high. I verified all 2675 keys rather than spot-checking: exact parity with English, zero placeholder mismatches, zero tag-structure mismatches, and array-valued keys (`tunnelTypes.*.diagram`, `nav.*.keywords`) intact in both type and length. That's better than most locale contributions.

Two things meant it couldn't land as-is, both my side rather than yours:

1. It targets `main`, but features here merge to `dev`, which is 85 English keys ahead — the whole plugin consent dialog, plugin update/integrity strings, and a few others. Merging to `main` would have shipped zh incomplete at the next release.
2. `keyParity.test.ts` only loads `fr` and `ru`, so the green suite never actually checked your files. Not something you could have known from the outside.

I've opened #80, which cherry-picks your commit unchanged — authorship preserved — onto `dev`, translates the 85 missing keys following the terminology you established, and extends the parity test to cover every locale in both directions so this can't silently drift again. I also normalized 44 ASCII colons to full-width `：` after Chinese text, and labelled the locale `简体中文` so a Traditional variant can be added later.

Closing in favour of #80. You're credited in the CHANGELOG. Thanks again — please do open more.

Issue #94, opened 2026-08-07
[BUG] Closing app take 2 seconds


Flash303:
## Description
Wen i try to close the application, it's so long

## Steps to Reproduce
1. Open the app
2. Click on the close button

## Screenshots / Video
[Youtube video](https://youtu.be/mlnFjHAY-tU)

## Environment Info
*Please tell us what device you are using so we can reproduce the bug.*

* **OS:** Windows 11
* **App Version:** On the vidéo v0.18.0, but also on any previous versions

kipavy:
Hi, its weird, I've never encounter this on windows 11 both on x64 and ARM... Your PR seem to change the closing behavior to a "suspend to tray" am I right ? I dont really want this for Voltius, closing should just close, without this issue you're facing, I'll look into it. Thanks

Flash303:
> Hi, its weird, I've never encounter this on windows 11 both on x64 and ARM... Your PR seem to change the closing behavior to a "suspend to tray" am I right ? I dont really want this for Voltius, closing should just close, without this issue you're facing, I'll look into it. Thanks

​Hello,

​This happens with the Gist sync plugin.
On the closing event, Voltius starts a sync with Gist before closing the app. So, when you click on the close button, you need to wait about 2 seconds before the app is really closed.

​This PR just hides the Tauri window; the sync is done in the background, and the app closes instantly.

​I'm also on Windows 11. If you want to try enabling the plugin, you can see if it's just my Windows installation or the app itself.

kipavy:
> > Hi, its weird, I've never encounter this on windows 11 both on x64 and ARM... Your PR seem to change the closing behavior to a "suspend to tray" am I right ? I dont really want this for Voltius, closing should just close, without this issue you're facing, I'll look into it. Thanks
> 
> ​Hello,
> 
> ​This happens with the Gist sync plugin. On the closing event, Voltius starts a sync with Gist before closing the app. So, when you click on the close button, you need to wait about 2 seconds before the app is really closed.
> 
> ​This PR just hides the Tauri window; the sync is done in the background, and the app closes instantly.
> 
> ​I'm also on Windows 11. If you want to try enabling the plugin, you can see if it's just my Windows installation or the app itself.

Oh alright, Thanks anyway for the PR, I'll look into it very soon

Flash303:
Fixed by #97

kipavy:
Expect a new release very soon, most likely today

PR #95, opened 2026-08-08
fix: instant hide the window and proces in background

Flash303:
This pull request introduces improvements to window management permissions and the quit handler behavior in the application. The main focus is on allowing the window to be hidden and ensuring the window is hidden before executing quit logic.

**Window management enhancements:**

* Added the `core:window:allow-hide` capability to the `default.json` configuration, enabling the application to programmatically hide the window.

**Quit handler improvements:**

* Updated the `ensureQuitHandler` function in `runtime.ts` to call `win.hide()` when a close is requested, ensuring the window is hidden before running quit callbacks.

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -10,6 +10,7 @@
     "core:window:allow-minimize",
     "core:window:allow-toggle-maximize",
     "core:window:allow-close",
+    "core:window:allow-hide",
     {
       "identifier": "opener:allow-open-url",
       "allow": [
--- a/src/plugins/runtime.ts
+++ b/src/plugins/runtime.ts
@@ -184,6 +184,7 @@ async function ensureQuitHandler() {
   const win = getCurrentWindow();
   await win.onCloseRequested(async (event) => {
     event.preventDefault();
+    win.hide().catch(() => {});
     const callbacks = [..._onBeforeQuit];
     await Promise.race([
       Promise.allSettled(callbacks.map((cb) => cb())),


Flash303:
Related to #94

kipavy:
Thanks @Flash303 — your diagnosis was right, and I could confirm it in the code: `ensureQuitHandler()` in `src/plugins/runtime.ts` preventDefaults the close and then awaits every plugin `onBeforeQuit` callback with the window still on screen. gist-sync pushes to GitHub in that hook, hence the ~2s. With no plugin registering a quit hook the handler is never installed at all, which is why I never hit this myself.

I've opened #97, which takes your approach (hide first, let the sync finish in the background) plus two safety fixes:

- Only hide when a quit hook is actually registered — otherwise close stays immediate as before.
- Arm the Rust-side exit up front with a fallback timer and call it from a `finally`, so a throwing hook or a failed `invoke` can't leave a hidden window over a live process.

It also adds unit tests for the handler, and I gated it on a real build under Xvfb/WebKitGTK with a test plugin whose quit hook runs ~2.4s: the window reports invisible from 303ms onward while the hook is still running, and the process exits 2.97s after the close — right after the hook, not on the fallback. With no hook registered, exit was ~240ms.

I don't have a Windows machine to check against, so if you get a chance to build #97 (or try the next dev build) with gist-sync enabled and confirm the close feels instant on your side, that would be great. I'll most likely close this PR in favour of #97, but the fix is yours — credited there.

kipavy:
Superseded by #97, which carries this fix with the extra safeguards. Thanks again @Flash303.

Flash303:
> Superseded by #97, which carries this fix with the extra safeguards. Thanks again @Flash303.

Np

PR #97, opened 2026-08-10
fix(plugins): hide the window before running quit hooks (#94)

### LINKED ISSUE #94: [BUG] Closing app take 2 seconds

kipavy:
Fixes #94. Based on @Flash303's diagnosis and patch in #95 — the approach here is theirs, with two extra safety fixes.

## Cause

`ensureQuitHandler()` (`src/plugins/runtime.ts`) intercepts `onCloseRequested`, calls `preventDefault()`, and then awaits every plugin `onBeforeQuit` callback (capped at 5s) before triggering the Rust-side exit. The window stays on screen for that entire wait. The gist-sync plugin pushes to GitHub in its quit hook, so closing the app looked frozen for ~2s. With no plugin registering a quit hook the handler is never installed at all, which is why this only reproduces with a plugin like gist-sync enabled.

## Change

- Hide the window before awaiting the hooks, so the wait is invisible and the app appears to close instantly. Requires the `core:window:allow-hide` capability (not part of `core:default`).
- Only hide when a quit hook is actually registered; with none, close immediately as before.
- Arm the Rust-side exit up front with a fallback timer and call it from a `finally` block. Without this, a throwing hook or an `invoke` that never lands would leave a hidden window over a live process — worse than a slow close.
- Wrap hook invocation in an async function so a synchronous throw becomes a rejection instead of escaping the handler.

## Tests

New `src/plugins/runtime.quitHandler.test.ts` covers: hide precedes the hook, no hide when there are zero hooks, exit still happens when a hook throws, and exit on the 5s cap with the fallback timer disarmed (`force_quit` invoked exactly once). `pnpm tsc --noEmit` clean; `pnpm vitest run src/plugins` green (73 files, 593 tests).

## Live verification

Debug build driven under Xvfb/WebKitGTK through tauri-driver, with a test plugin whose quit hook runs for ~2.4s:

- Close fired, hook ran, `plugin:window|is_visible` sampled from inside the hook: `false` at every sample from 303ms onward through 2429ms — the window is gone while the sync is still running.
- Process exited 2.97s after the close was fired, i.e. right after the hook finished, not on the 6s fallback.
- With the hook unregistered, the process exited 240ms after the close was fired (that figure includes the harness round-trip), and no hide occurred.

Not verified on Windows — no Windows machine here. The mechanism is platform-independent, but @Flash303 confirmed the original symptom and fix on Windows 11.

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -10,6 +10,7 @@
     "core:window:allow-minimize",
     "core:window:allow-toggle-maximize",
     "core:window:allow-close",
+    "core:window:allow-hide",
     {
       "identifier": "opener:allow-open-url",
       "allow": [
--- a/src/plugins/runtime.quitHandler.test.ts
+++ b/src/plugins/runtime.quitHandler.test.ts
@@ -1,0 +1,101 @@
+import { describe, test, expect, beforeEach, afterEach, vi } from "vitest";
+import { loadPlugin, unloadPlugin } from "./runtime";
+import type { PluginAPI, PluginManifest } from "./api";
+
+type CloseHandler = (event: { preventDefault: () => void }) => unknown;
+
+const { closeHandlers, hide, invoke } = vi.hoisted(() => ({
+  closeHandlers: [] as CloseHandler[],
+  hide: vi.fn(() => Promise.resolve()),
+  invoke: vi.fn(() => Promise.resolve()),
+}));
+
+vi.mock("@tauri-apps/api/window", () => ({
+  getCurrentWindow: () => ({
+    hide,
+    onCloseRequested: (cb: CloseHandler) => {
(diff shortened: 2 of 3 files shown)


Flash303:
> Not verified on Windows — no Windows machine here. The mechanism is platform-independent, but @Flash303 confirmed the original symptom and fix on Windows 11.

Works well on Windows, thanks and waiting for the release !!

PR #113, opened 2026-08-13
feat(mcp): team and sharing verbs (P7, 64 -> 74)

kipavy:
## What

MCP max-surface **P7 — team and sharing**. Ten new verbs, taking the tool surface from **64 to 74**:

`team_list`, `member_list`, `vault_key_status`, `member_invite`, `member_remove`, `member_set_role`, `list_shared_sessions`, `share_session`, `unshare_session`, `handoff_control`

Four new permissions (`team:read`, `team:write`, `sharing:read`, `sharing:write`), six new audit actions, and the `settings.integrations.mcp.sub` copy widened in all four locales.

## Shape

Two pure domain modules (`src/plugins/domains/team.ts`, `sharing.ts`) hold the logic behind injected `*Ports` interfaces; two tool modules (`src/plugins/toolSurface/tools/team.ts`, `sharing.ts`) own the zod schemas, permissions, approval gate and audit. `runtime.ts` wires the real stores. Nothing under `toolSurface/` imports a store.

## The gates that bound an agent here

- **Server role bits** bound every membership write. `PERM_MANAGE_MEMBERS` / `PERM_INVITE_MEMBERS` are enforced server-side, so an agent can do nothing through MCP its user could not do by hand. Refusals are surfaced, never swallowed.
- **Session ownership** bounds the sharing verbs, checked before *and* after the approval gate. A session the user opened has no ownership row, and that absence is what keeps an agent out of the user's own terminals. `list_shared_sessions` reports `agentOwned` from the pure `owned.has` — never `mayAct`, which can adopt.
- **The broadcast refusal.** `routeInputBytes` fans typed input to every session in an active broadcast split tab, so sharing such a session would stream the user's own keystrokes to remote participants. `share_session` refuses, ahead of both the approval card and the audit row. The predicate is defined once, in `layoutStore`, and is now shared by the terminal input path, snippet broadcast, and this refusal — it had drifted into three copies.
- **`handoff_control` only approves a pending request.** Control is never handed to a participant who has not asked for it in that moment.
- **No invite links.** That path mints a bearer token granting a live terminal to whoever holds the string; it is unreachable from this surface, and no token appears in any result or audit row.
- **Audit rows name the counterparty** (vault ids on a share, the target user on a grant or a removal) but carry no invitee email, session key, or token.

`vault_key_status` reports the keyless window — a member who *can* be given a wrapped vault key but has not been — rather than hiding it behind a successful invite. That window is real, observed behaviour.

## Also touched outside the new files

`useTerminal.ts`, `services/snippets.ts`, `ShareMenu.tsx` (shared-helper extractions), `toolSurface/helpers.ts` + `refusal.ts` (a returned refusal now passes through `objectOp` instead of nesting), `gatedPermissions.ts` (the four new permissions had no consent tier and rendered as unknown strings), and the plugin-types generator.

## Review

Per-task review after each of nine tasks, plus a whole-branch review. No Critical findings; the whole-branch pass caught five Important correctness bugs that mocked tests structurally could not — among them `member_set_role` taking a `roleId` no read verb exposed while removing every existing role first, which made it both uncallable and destructive.

## ⚠️ Before release

- **`voltius-server` must be rebuilt** with the audit whitelist change (server repo `2ff7248`), or all six new team audit rows are 400ed and **silently dropped**.
- **The live two-account E2E gate has not been run.** It is this phase's real gate. Three things only it can prove: the keyless window on a real invitee; that `controlRequester` actually populates over the WebSocket (if it does not, `handoff_control` is permanently refused and no mock can tell); and that the audit rows land server-side.
- **A fluent read of the fr/ru/zh** `mcp.sub` strings is owed, plus P6's leftover `panes.header.mcpTooltipDisconnected` in ru/zh.


kipavy:
## Audit scope fixed, and re-verified live

The live two-account gate found that the three membership verbs' audit rows never reached the **team server** — they were filed only in the on-device sink, so a team admin could see *that* membership changed (the server's own `member.*` rows) but not that an **agent** did it.

Cause chain: `deriveScope` returned null for those verbs → the consumer used the literal `"mcp"` as scope → `audit.record` matched no connection → context fell back to `{ kind: "local", vaultId: "personal" }` → `reportPluginAuditEvent` forwards only on `kind === "team"`.

Fixed in `ceb0bd2` (+ `b0c079d`, comments):
- `deriveScope` returns the **team id** for `member_invite` / `member_remove` / `member_set_role`, failing closed to null without one. The scope is also the approval allowlist bucket, so a grant is keyed to one team — the right grain for a membership change.
- `audit.record` promotes a scope to team context **only** when it resolves to `kind === "team"`. Everything else keeps the exact `{ local, personal }` fallback — the resolver returns `vaultId: <id>` for an unknown id, and letting that through would file rows in a local bucket `audit.query` never reads.

**Live proof** (fresh two-account stack, real `voltius mcp` client). The server's `audit_logs` now contains:

```
agent.member_invited      | {via: mcp, tool: member_invite,   teamId, userId, approval, plugin_id}
agent.member_role_changed | {via: mcp, tool: member_set_role, teamId, userId, approval, plugin_id}
```

No email, no token, no key material. Before the fix that table held only the server's own rows, so Task 1's whitelist addition is now exercised for real.

Sharing verbs deliberately unchanged — their scope stays the shared session's connection, which is the right approval grain; re-scoping them to a shared-to vault is a separate call.

PR #115, opened 2026-08-14
feat: terminal line-height & cursor settings, fix cursor blink under tmux

xtac:
## What

**Terminal appearance settings** (`c837652d`)
- Themes gain an optional `terminalLineHeight` (xterm `lineHeight` multiplier, default 1), editable in the theme creator next to the terminal font size. Optional field, so existing saved/custom themes are unaffected.
- Settings → Terminal gains two new controls:
  - **Cursor style** — bar / block / underline (default: bar, the previous hardcoded value)
  - **Cursor blink** — on/off toggle (default: on, the previous hardcoded value)
- Both apply live to already-open terminals, not just new ones.
- The live theme-apply block that was copied across `useTerminal`, the theme-preview handler, and `MultiplayerTerminalView` is unified into a shared `applyTerminalTheme()` helper.
- Labels translated for en/fr/ru/zh; "cursor" added to the settings search keywords.

**Cursor blink fix for persistent sessions** (`d4cef8d6`)
- With persistent sessions enabled, SSH sessions are wrapped in tmux/screen on the host. Every multiplexer redraw emits xterm-256color's `cnorm` (`\E[?12l\E[?25h`), and xterm.js maps DECRST 12 straight onto `cursorBlink = false` — so the cursor silently stopped blinking in every wrapped SSH session, while local sessions kept blinking.
- The generated tmux/screen configs now override `cnorm`/`ve` to plain `\E[?25h`. Remote apps that explicitly request a blinking cursor (cvvis / DECSCUSR) still work; only the implicit blink-off on redraw is dropped.
- Note: an existing tmux server keeps its old config — the fix applies once the server restarts (`tmux -L voltius kill-server` or host reboot).


kipavy:
Merged into `dev` as 3f08ce7 + 072516c. Rebased instead of button-merged (this branch was 139 commits back, before `settingsManifest.ts` existed), so GitHub won't auto-close it — closing by hand. Your authorship is kept.

Nice catch on `cnorm`. Checked it at the byte level on Alpine/tmux 3.6b: stock config emits `ESC[?12l`, with the override it doesn't.

Follow-ups added in 626a1d4:
- line height clamped before reaching xterm (empty/mid-typed field gives 0 or NaN → zero-height rows); clamped on read and on blur, not per keystroke
- `terminal.cursorStyle` added to `explicitDefs()` so MCP/plugins can read+write it (the blink toggle was already covered via `TOGGLE_DEFS`)
- cursor style added to app-settings export/import so it actually syncs
- shared `subscribeTerminalCursor` (style+blink subs were duplicated in `useTerminal` and `MultiplayerTerminalView`)
- `cursor` search keyword for fr/ru/zh
- tests in `src/utils/terminalTheme.test.ts`

Live-gated against a real SSH host: line height 1.8 took the PTY 39 → 22 rows, and a fresh persistent session kept blinking across a `clear` while stock tmux froze the cursor.

Caveat: `?12l` is one-way in xterm.js. Nothing re-enables blink, so besides `tmux -L voltius kill-server` an already-affected tab needs reopening.

PR #117, opened 2026-08-14
fix: prevent window from closing if reopened during shutdown sync

Flash303:
The problem:
[https://youtu.be/MN6DXxceRuI](https://youtu.be/MN6DXxceRuI)

When I close the app, the window hides and the sync runs in the background.
However, if you reopen the app before the sync is finished, the window closes once it completes.

This PR fixes this bug; the app will no longer close if the window is shown.

### DIFF
--- a/src/plugins/runtime.quitHandler.test.ts
+++ b/src/plugins/runtime.quitHandler.test.ts
@@ -4,15 +4,17 @@ import type { PluginAPI, PluginManifest } from "./api";
 
 type CloseHandler = (event: { preventDefault: () => void }) => unknown;
 
-const { closeHandlers, hide, invoke } = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


Flash303:
Related to #97

kipavy:
Thanks — real bug, and the cause (single-instance `window.show()` landing mid-wait) is the right one. I pushed 727fcb2 on top with three hardening fixes:

- `win.hide()` is now awaited (bounded to 1s). Unawaited, a failed hide was indistinguishable from a reopened window, so `isVisible()` would report true and the app could never close.
- The 6s fallback is disarmed before the visibility check, and the check is bounded too, defaulting to exiting — so neither a slow `isVisible()` nor a hung one can strand the quit.
- Added tests for the reopened / rejected / hung / failed-hide paths.

Behaviour you fixed is unchanged.

PR #141, opened 2026-08-17
fix(vault): stop the vault becoming unreadable, and give users a way back when it does

kipavy:
Closes the four vault follow-ups left after #139, plus two prevention fixes and one gap found while verifying them. None had an issue. Ordered by what they protect against, not by discovery.

## Prevention — stop the vault becoming unreadable

**1. `secrets.enc` was written non-atomically.** `save()` overwrote the file in place, so a crash, kill or full disk mid-write left a truncated file that no key decrypts and nothing had backed up. It now stages into a sibling `.tmp`, fsyncs, and renames. Every user was exposed to this on every password save.

**2. The legacy-account migration re-encrypted before the new key was recoverable.** `migrateToWrappedUserSecrets` rekeyed `secrets.enc` from the kek to a freshly generated dek and *then* uploaded that dek. A failed upload fell back to the kek (`account.ts:739`, `:752`) with the file already keyed to a dek held nowhere — permanently unreadable, no backup, no warning. The order is now swapped: a failed upload leaves the file untouched, and a failed rekey degrades to a pre-split device, which `login` already handles by keeping the kek.

Prod has **0 of 194** accounts still un-migrated, so this can now only fire on a self-hosted server with legacy accounts. Fixed anyway — the failure mode is total.

## Recovery — a way back when it does happen

**3. Nothing in the app could restore a quarantined vault.** `secrets_quarantine` keeps three `.bak` files that no screen listed; `grep -rn '\.bak' src/` returned nothing. Someone who took "set aside and re-download" and got a stale or empty cloud blob had their secrets on disk with no in-app path back.

- `secrets_backups` lists them newest-first with stamp and size.
- `secrets_restore` **copies** the chosen backup into place after setting the current vault aside, so a restore deletes nothing and a wrong choice is undone by restoring another. It prunes nothing either: retention must never be able to delete the file a restore is reading.
- The file name is validated against `<vault>.<digits>.bak` before any path is built from it — it arrives from the frontend.
- One `VaultBackups` component, mounted on `AuthPage`'s recovery screen (hidden when empty) and in Settings › Account, where the confirmation says the current vault is readable and will be displaced.

**4. A no-password account with an unreadable vault opened the app anyway.** `autoLogin`'s OS-keychain branch installed the key without ever testing it against `secrets.enc` — the password branch declines, this one did not. So the splash succeeded, the main UI painted, and the failure surfaced at the first secret read, *inside* the app, where the only exits were an unlock prompt for a password that does not exist and a reload.

It now verifies like the password branch, and `autoLogin` returns `"ok" | "declined" | "vault-unreadable"` so `SplashScreen` can tell the two failures apart: `declined` keeps the unlock prompt, `vault-unreadable` goes to the recovery screen with set-aside and the backup list. That screen drops the retry link and the re-download wording when there is no password to retry and no cloud copy to re-download.

This is also the answer to "how would a user ever see an in-app vault error": on a master-password account they cannot, which is why items 5 and 6 below are defence in depth everywhere except here.

## Reporting — stop a vault failure looking like something else

**5. Form editors gave no signal that a secret could not be read.** ~50 sites wrapped `getSecret` in `.catch(() => null)`, so an unreadable vault rendered a password field indistinguishable from one that was never saved. An untouched field is **not** overwritten — `buildSubmit` sends `password: passwordDirty.current ? password : null` and `saveHostFromForm` only writes when non-null — so the risk is a user typing a guess over a real secret, not a silent wipe.

One `useStoredSecrets` hook now loads a form's secrets, keeps "not stored" apart from "the vault could not answer", and drives a note on ConnectionForm, IdentityForm and KeyForm. The duplicate and copy-to-vault flows on the keychain and hosts pages drop the same swallow (23 sites): a vault failure there produced a credential-less copy in silence, and every caller already surfaces the error.

**6. Rust's locked store carried no code.** `secrets_get/set/delete` answer with a bare `Secrets store is locked`, which reached the generic error panel instead of the one offering to unlock. The three now go through one helper that maps it to `VaultLockedError` and clears the module's `unlocked` flag, which by definition has drifted from the store.

**7. SidePane's SFTP overlay printed the raw message.** Its `ConnectionOverlay` was `status="connecting"` only. The vault code now rides on the error phase and the overlay renders `VaultErrorPanel`. The host icon/name/subtitle both branches need is one helper rather than two copies.

## Verification

- **3642 TS tests across 478 files**, `npx tsc --noEmit` clean, **323 Rust tests**, `cargo fmt --check` clean.
- Every fix was shown failing first and proven red by reverting the fix, not by assuming — including the non-atomic write, the migration ordering (rekey ran at step 17, upload at 19), the backup-name traversal guard, copy-not-move on restore, and the no-password verification.
- Two Rust tests take an injected clock so retention and naming fail deterministically rather than depending on how many operations share a millisecond.

**Driven in the live app:**
- Restoring a backup produced a `secrets.enc` byte-identical to the source (`md5 7eb98c9a…`) with **all three backups still on disk**, then sent the app to the unlock screen — the restored file answers to whichever key encrypted it.
- Saving a password wrote `secrets.enc` through the staged path with no `.tmp` left behind.
- The whole no-password recovery loop, unpatched: create the account, corrupt `secrets.enc`, relaunch → recovery screen (before this branch the app opened normally); "Set aside and start fresh" → `secrets.enc.1786994429902.bak`; corrupt again → the same screen now lists that backup with a Restore button.

## Not covered

- The **cloud re-auth** variant of the recovery screen (`vaultUnreadableBody`, the wrong-key wording) is still unit-covered only: it needs `login`'s re-auth branch and a reachable server.
- fr/ru/zh strings here are unreviewed by a native speaker.
- The migration fix cannot be exercised against prod — there is nothing left to migrate.
- Cosmetic: in the narrow auth column a backup filename ellipsises and its date/size wraps to two lines.
- The system-toast envelope `{ source: { kind: "plugin", id: "system", name: "Voltius" }, … }` now has 18 copies including this one. Unifying it touches modules unrelated to these items, so it is left consistent and flagged rather than widening this branch.


kipavy:
Second review pass on this branch found nine issues; all are now fixed in the three commits above, each with a regression test that was checked to fail without its fix.

The two that mattered:

- **Retention could delete backups a recovery run never tried.** Restore never prunes, but quarantine still did, and each restore *adds* a backup (the vault it displaces). With b1..b3 on disk, restoring b3 and then pressing "Set aside and start fresh" — the button sitting directly above the backup list — pruned to three and took b1 and b2. A restore now keeps what it displaces as `<name>.<stamp>.displaced.bak`, and retention stands down entirely while any displaced backup exists.
- **The legacy-vault migration rekeyed a locked store.** `secrets_rekey` requires an unlocked store and nothing on the `login` path unlocked one, so a cold legacy login uploaded `wrapped_user_secrets`, threw on the rekey, and had it swallowed: the account was marked migrated server-side while the device kept a kek-encrypted vault, with no dek in the session for team crypto and no sync push. The unlock now runs before the upload.

Also: restore copies the current vault aside before renaming, so a crash mid-restore can no longer leave *no* vault (which the next launch reads as a first launch); the staged write inherits the vault's mode instead of widening a `chmod 600` vault to 644; `secrets_restore` validates the backup name before locking the store; read errors are no longer reported as "wrong key", which had been aiming transient failures at the destructive recovery screen; the backup list shows real byte sizes instead of rounding a truncated file up to "1 kB"; and `VaultUnavailableNote` + `ErrorBanner` take their colours from status tokens rather than a hardcoded dark palette that rendered dark-on-dark in the light theme.

**One deliberate behaviour change worth knowing:** once a restore-displaced backup exists, retention does not prune again until that file is gone. Backups are only ever created by a quarantine or a restore — ordinary saves never add one — so growth stays bounded in practice.

Verified: `cargo test` 330 passed, `cargo fmt --check` clean, `vitest run` 3649 passed across 478 files, `tsc --noEmit` clean, `pnpm build` succeeds.

PR #177, opened 2026-08-24
fix: update macOS quarantine guidance

### LINKED ISSUE #176: Remove obsolete --no-quarantine instructions from Homebrew cask and GitHub docs
### Description

Voltius currently recommends the Homebrew option `--no-quarantine` in the generated cask caveat and in GitHub documentation:

```sh
brew install --cask --no-quarantine voltiusapp/voltius/voltius
```

On Homebrew 6.0.19, `brew install` rejects this command with:

```text
Error: invalid option: --no-quarantine
```

The normal installation succeeds, but users following the displayed or documented workaround receive an error.

### Steps to reproduce

1. Install the cask:

   ```sh
   brew install --cask voltiusapp/voltius/voltius
   ```

2. Run the suggested `--no-quarantine` command.
3. Observe that Homebrew prints its usage and exits with `Error: invalid option: --no-quarantine`.

### Affected locations

- [ ] `VoltiusApp/voltius`: update the macOS/Homebrew section in `README.md`.
- [ ] `VoltiusApp/voltius`: update `scripts/gen-homebrew-cask.sh`, which generates future tap releases.
- [ ] `VoltiusApp/homebrew-voltius`: update `README.md`.
- [ ] `VoltiusApp/homebrew-voltius`: update the currently published `Casks/voltius.rb` caveat so existing users immediately receive correct guidance.

### Suggested change

Remove every recommendation to pass `--no-quarantine` to Homebrew. Keep Control-click → **Open** as the preferred first-launch workflow.

If a command-line workaround is retained, document it as a post-install command and explain that it removes Gatekeeper quarantine metadata:

```sh
xattr -dr com.apple.quarantine /Applications/Voltius.app
```

The direct `.dmg` guidance in the main README should use the same wording and command so the Homebrew and direct-download instructions remain consistent.

It would also be useful to add a small assertion around the cask generator output so `--no-quarantine` cannot be reintroduced by a future release.

### Environment

- macOS
- Homebrew 6.0.19
- Voltius cask 0.30.0

MrDoudou:
## Summary

- replace the removed Homebrew `--no-quarantine` option with a supported post-install `xattr` command
- keep Control-click → Open as the preferred Gatekeeper workflow
- make the direct `.dmg` instructions use the same quarantine guidance
- prevent generated casks from reintroducing `--no-quarantine`

## Validation

- `bash -n scripts/gen-homebrew-cask.sh`
- generated and validated both tap and `--core` cask variants for `v0.30.0`
- verified the tap variant matches the corrected published cask
- `ruby -c` on both generated variants
- `brew style` on the generated tap cask
- `git diff --check`

Fixes #176


kipavy:
Thanks — this is the right fix. I retargeted it from `main` to `dev`, which is where feature and fix branches land here; `main` only moves on a release merge. Nothing else needed from you.

Keeping `--core` was the better call than collapsing the two variants: core is hostile to caveats that talk the user out of Gatekeeper, not only to `--no-quarantine`, so the `xattr` line belongs in the tap variant alone.

The rest of the issue's surfaces are covered alongside this:

- VoltiusApp/homebrew-voltius#2 — the published cask caveat and tap README, so existing users stop seeing the broken hint before the next release regenerates the cask. Its cask is byte-identical to `gen-homebrew-cask.sh v0.30.0` on this branch.
- VoltiusApp/web#15 — landing page download card
- VoltiusApp/docs#17 — docs site install page
- #178 — one stale paragraph in `packaging/README.md` that still describes the `--no-quarantine` hint; merge after this.

PR #189, opened 2026-08-26
fix(team): carry a vault's existing contents into the team vault it becomes

### LINKED ISSUE #188: Turning a private vault into a team vault leaves its existing hosts behind
Found while live-testing #70 (PR #186) with two real accounts against an isolated server.

## What happens

Everything a vault already contained stays local to the owner when the vault is converted to a team vault. The owner keeps seeing it; the team vault is empty for everyone else, forever, with no warning.

## Why

`convertVaultToTeam` (`src/services/vaultConvert.ts`) creates the team, links the vault, and initialises the key — and stops there:

```ts
const team = await useTeamStore.getState().createTeam(vaultName);
useVaultStore.getState().setVaultTeamId(vaultId, team.id);
const { initTeamVaultKey } = await import("@/services/teamVaultSync");
await initTeamVaultKey(team.id, []);
markTeamVaultLoadedAfterLocalActivation(team.id, useTeamVaultStateStore.getState());
```

Nothing walks the vault's connections / identities / keys / folders / snippets and writes them through `saveTeamVaultObject`. They keep living in the local stores under the old vault id, so the owner's UI still lists them under the (now team) vault while `team_vault_objects` stays empty.

Writes made *after* the conversion propagate normally — this is only the pre-existing contents.

## Reproduction

1. Create a private vault `Ops`, add one host to it.
2. Members → Invite → "Turn into a team vault".
3. `select count(*) from team_vault_objects` → **0**. The owner still sees "Ops · 1 host".
4. Invite a second account; it joins, unlocks, and sees "No hosts yet".
5. Owner adds a second host → `team_vault_objects` = 1 → the invitee sees that one, and only that one.

## Why it matters beyond the missing data

The whole point of the conversion consent sheet is that the user understands what sharing means. It currently promises the opposite of what happens: the user shares "Ops", their teammate gets an empty vault, and the difference is invisible from the owner's side because their own copy looks complete.

It also makes the #70 first-access experience land on an empty vault in the most common setup path.

## Options

- Migrate on conversion: push every object in the vault through `saveTeamVaultObject` as part of `convertVaultToTeam`, and fail the conversion visibly if the upload fails rather than half-converting.
- Or, if migration is deliberately not wanted, say so in the consent sheet ("existing hosts stay private; only what you add afterwards is shared") — but that reading seems worse than the bug.

Related: #68 (the share sheet that owns this surface), #70.

kipavy:
Fixes #188.

## The bug

`convertVaultToTeam` created the team, linked the vault, initialised the key — and stopped. Nothing walked the vault's connections / identities / keys / folders / snippets / port-forwarding rules through `saveTeamVaultObject`, so they kept living in the local stores under the old vault id. `team_vault_objects` stayed empty: the owner's UI still listed "Ops · 1 host" while every teammate saw an empty vault, permanently and with no warning. Only writes made *after* the conversion propagated.

The Settings → Vaults conversion path (`PrivateVaultMembersPanel`) already did the migration, with its own copy of the sequence. The share sheet's path never called it.

## The fix

**One migration, one conversion.** `migrateVaultToTeam` moves out of `VaultsSection.tsx` into `src/services/vaultTeamMigration.ts` and becomes table-driven over the seven object kinds — the collect / upload / publish / delete / reload blocks were written out per kind five times over. Both conversion paths now call one `createTeamVaultFromVault`, so the share sheet cannot drift from Settings again.

**All-or-nothing.** The uploads run first and as a group. Until they have all succeeded nothing local has been touched, so a failure unlinks the vault, drops the team locally and `DELETE`s it on the server instead of leaving a vault that looks complete to its owner and empty to everyone else. Past that commit point the remaining steps (secret backfill, local deletion, store reload) are best-effort and never throw — the objects already live in the team vault by then.

**Consent sheet.** It now says the existing contents move, which is what the conversion does.

### Incidental

- `teamService.deleteTeam` — the make-private path had this `fetch` inline; the rollback needed the same call.
- `reloadLocalVaultObjectStores` — the same seven-store reload, previously written twice.
- `teamStore.removeTeam` clears `activeTeamId` when it pointed at the removed team.

## Verification

- `tsc --noEmit` clean; full suite 3952 passed / 514 files.
- 11 new/changed tests. The four `vaultConvert` tests all fail with the `migrateVaultToTeam` call commented out, so they hold the bug down.
- Not yet live-run against two real accounts — the repro in #188 is the check worth doing before merge.


kipavy:
Live-verified against an isolated server with two real accounts (`voltius-e2e-*` stack, two keychain-isolated app containers on this branch).

The #188 repro, step by step:

1. Alice registers, creates private vault `Ops`, adds host `prod-web` (voltius@ssh-host-1). Baseline: `select count(*) from team_vault_objects` → **0**.
2. Members → Invite → the consent sheet now reads *"…Everything already in it moves across, so the people you add see it too."* → **Turn into a team vault**.
3. `select object_type, name from team_vault_objects` → **1 row, `connection | prod-web`**. On `dev` this is 0 — that is the bug.
4. Alice invites Bob by handle; Bob accepts; `team_members` = 2; Alice wraps the key (`Vault keys upserted key_count=1`, `team_vault_keys` = 2) and Bob's client fetches it (`Vault key fetched user_id=<bob>`).
5. Bob opens `Ops` → **`prod-web` renders**, `voltius@ssh-host-1`. Previously "No hosts yet".
6. Alice's own view still lists exactly one `prod-web` — the local row is gone and the team object is the only copy, so no duplicate.

Not exercised live: the rollback branch (no way to fail an upload on demand from the UI) — it is covered by the unit tests instead.

Incidental, not a code issue: two app containers share the host's keyutils quota for uid 1001, so the second registration first failed with `Keychain write error: Platform failure: QuotaExceeded`. Raising `kernel.keys.maxbytes` for the run fixed it; restored afterwards.

PR #209, opened 2026-09-07
fix(terminal): let Ctrl+G reach the shell when search is closed

### LINKED ISSUE #208: [BUG] Ctrl+G never reaches the shell — cannot abort a Ctrl+R reverse-i-search
## Description

`Ctrl+G` never reaches the shell. The key is intercepted unconditionally in the
frontend, so the PTY never receives `0x07` (BEL) — which is the byte readline's
`abort` function needs. The practical effect is that a `Ctrl+R` reverse-i-search
in bash cannot be aborted: `Ctrl+R` works, then `Ctrl+G` does nothing at all.

Any other shell/TUI use of `Ctrl+G` is affected the same way (readline `abort`,
emacs `keyboard-quit`, nano's help toggle, etc.).

## Steps to Reproduce

1. Open a terminal session (a local shell reproduces it; no SSH needed).
2. Press `Ctrl+R` and type a few characters — the `(reverse-i-search)` prompt appears.
3. Press `Ctrl+G`.
4. Nothing happens. The search prompt stays up and the original line is not restored.

## Expected Behavior

`Ctrl+G` aborts the incremental search and restores the pre-search line, as it
does in any other terminal emulator.

## Environment Info

* **OS:** Fedora 44 (Linux 7.1.10-200.fc44.x86_64)
* **App Version:** v0.31.1

I only tested on Fedora, but the code path below has no platform branch, so the
Windows and macOS builds are very likely affected too.

### Cause

`src/hooks/useTerminal.ts:898-907`, in the `attachCustomKeyEventHandler` callback:

```ts
if (e.ctrlKey && !e.altKey && (e.key === "g" || e.key === "G")) {
  if (e.type === "keydown") {
    const ctrl = getTerminalSearchController(sessionId);
    if (ctrl?.getSnapshot().open) {
      if (e.shiftKey) ctrl.prev();
      else ctrl.next();
    }
  }
  return false;
}
```

The `return false` is outside the `open` check. When the search widget is closed
the branch does no work and still tells xterm to skip the key entirely — so no
byte is encoded and nothing is written to the PTY. `src/hooks/useKeyboard.ts:66-81`
holds the matching window-level handler, which `preventDefault()`s on every
`Ctrl+G`.

### Origin, and the tradeoff I'd like your call on

This came in with 8eda22a9 "fix ctrl+g opening native webview search" (2026-05-24,
first released in v0.1.50), so it isn't a recent regression. That commit was
fixing a real problem — WebKitGTK binds `Ctrl+G` to find-next — and swallowing the
key is what stops that dialog.

The fix I have in mind is to intercept only when the search widget is open and
`return true` otherwise. xterm's own keyboard evaluation marks `ctrl`+letter as
`cancel: true` and calls `preventDefault()` itself, so the native find-next should
stay suppressed while `0x07` still reaches the PTY. That is a change to how your
earlier fix works, though, so I'd rather ask than assume — if you'd prefer a
different approach (an explicit `preventDefault()` plus writing `\x07` to the PTY
by hand, or gating on platform), I'm happy to do it that way.

Also worth noting: `Ctrl+G` is not a registered entry in `shortcutStore.ts` (only
`terminal-search` / `Ctrl+F` is), so there's no way for a user to rebind their way
out of this from Settings. The only workaround is a second abort binding in
`~/.inputrc`:

```
"\C-x\C-g": abort
```

I'm happy to open a PR against `dev` with a regression test — the xterm stub in
`src/hooks/useTerminal.clipboard.test.tsx` already captures the custom key handler,
so asserting the return value for open/closed search is straightforward.

tiepduong7c9:
Fixes #208.

## Problem

`Ctrl+G` never reached the PTY, so `bash` could not abort a `Ctrl+R`
reverse-i-search — readline's `abort` needs `^G` (0x07) and the byte was never
sent. `Ctrl+R` worked, then the key was simply dead, which is what made it look
like a `Ctrl+G`-specific fault. Every other shell/TUI use of the key went the
same way: emacs `keyboard-quit`, nano's help toggle, and so on.

`useTerminal.ts`'s custom key handler returned `false` for `Ctrl+G`
unconditionally — the `return` sat outside the check for an open search widget:

```ts
if (e.ctrlKey && !e.altKey && (e.key === "g" || e.key === "G")) {
  if (e.type === "keydown") {
    const ctrl = getTerminalSearchController(sessionId);
    if (ctrl?.getSnapshot().open) { /* next / prev */ }
  }
  return false;   // <- reached whether or not the widget is open
}
```

Returning `false` makes xterm skip the key entirely, encoding nothing, so with
the widget closed the branch did no work and still swallowed the key. Reported
on Fedora 44 / v0.31.1, but the path has no platform branch.

## Fix

The chord is now only claimed while the widget is open. On the pass-through
xterm marks `ctrl`+letter `cancel: true` and calls `preventDefault()` itself, so
the native webview find-next — the reason 8eda22a9 blocked the key, and not a
problem I wanted to reintroduce — stays suppressed.

Measured against the real `@xterm/xterm@6.0.0` in a browser, driving actual
`Control+g` keystrokes rather than synthetic ones:

| search widget | `onData` | event ends | reaches window handler |
| --- | --- | --- | --- |
| closed | `[0x07]` | `defaultPrevented: true` | no — xterm stops propagation |
| open | nothing | `defaultPrevented: false` | yes — `useKeyboard` cancels it there |

So both paths keep the native dialog shut, by different routes: xterm's own
cancel on the shell path, and the existing window handler on the widget path.

The reporter has also confirmed it by hand on the Fedora 44 / WebKitGTK build
this was found on: `Ctrl+G` aborts a reverse-i-search again, and the `Ctrl+F`
widget still walks its hits.

## Tests

`src/hooks/useTerminal.ctrlG.test.tsx` covers the handler's verdict for closed
widget, open widget (including that `keyup` stays consumed without moving the
hit again), and closed-again. Two of its three cases fail against the code this
replaces.

`pnpm test` — 522 files / 3991 tests pass. `pnpm build` clean. No Rust touched.

## The first commit

The two existing tests that mount a terminal each carried their own copy of the
same ~35-line xterm stub, and the regression test would have been a third. Per
the DRY note in `CLAUDE.md` I extracted one fixture instead of adding a copy.
It's a separate commit and drops cleanly if you would rather review a one-file
diff — the fix itself does not depend on it beyond using the fixture.

### DIFF
--- a/src/hooks/__fixtures__/fakeXterm.ts
+++ b/src/hooks/__fixtures__/fakeXterm.ts
@@ -1,0 +1,108 @@
+/**
+ * Shared xterm doubles for the useTerminal tests.
+ *
+ * Every one of those tests needs the same stub of the slice of the Terminal and
+ * addon surface useTerminal touches, so it lives here once: a new xterm API has
+ * to be taught to a single fake rather than to each test file that mounts a
+ * terminal.
+ *
+ * `vi.mock` factories are hoisted per file, so a test still declares its own
+ * mocks — they just point at these classes:
+ *
+ *   vi.mock("@xterm/xterm", async () => ({
+ *     Terminal: (await import("@/hooks/__fixtures__/fakeXterm")).FakeTerminal,
+ *   }));
+ *
+ * The factory has to do that import itself — vi.mock is hoisted above any
+ * top-level binding the file might otherwise share.
+ */
+
+/**
+ * Handlers registered through `attachCustomKeyEventHandler`, keyed by terminal
+ * id. Calling one is how a test delivers a key to the terminal that owns it.
+ */
+export const keyHandlers = new Map<string, (e: KeyboardEvent) => boolean>();
+
+let seq = 0;
+
+/** Terminal ids run `term-1`, `term-2`, ... in creation order. */
+export class FakeTerminal {
+  id = `term-${(seq += 1)}`;
+  element: HTMLElement | null = null;
+  options: Record<string, unknown> = {};
+  cols = 80;
+  rows = 24;
+  modes = { applicationCursorKeysMode: false, mouseTrackingMode: "none" };
+  buffer = {
+    active: { length: 0, viewportY: 0, baseY: 0, cursorY: 0, type: "normal", getLine: () => null },
+    onBufferChange: () => ({ dispose() {} }),
+  };
+  open(container: HTMLElement) {
+    this.element = document.createElement("div");
+    container.appendChild(this.element);
+  }
+  parser = { registerOscHandler: () => ({ dispose() {} }) };
+  loadAddon() {}
+  write() {}
+  focus() {}
+  getSelection() { return ""; }
+  dispose() {}
+  attachCustomKeyEventHandler(fn: (e: KeyboardEvent) => boolean) { keyHandlers.set(this.id, fn); }
+  attachCustomWheelEventHandler() {}
+  onData() { return { dispose() {} }; }
+  onBinary() { return { dispose() {} }; }
+  onResize() { return { dispose() {} }; }
+  onScroll() { return { dispose() {} }; }
+  onLineFeed() { return { dispose() {} }; }
+  onRender() { return { dispose() {} }; }
+  onWriteParsed() { return { dispose() {} }; }
+  registerLinkProvider() { return { dispose() {} }; }
+  registerDecoration() { return null; }
+}
+
+export class FakeFitAddon {
+  fit() {}
+  proposeDimensions() { return { cols: 80, rows: 24 }; }
+}
+
+export class FakeWebglAddon {
+  onContextLoss() { return { dispose() {} }; }
+  dispose() {}
+}
+
+export class FakeWebLinksAddon {}
+
+/**
+ * Records searches instead of running them — the real addon reports hits back
+ * through `onDidChangeResults`, which a stub cannot do, so `searches` is the
+ * only observable that a find actually ran.
+ */
+export class FakeSearchAddon {
+  static instances: FakeSearchAddon[] = [];
+  searches: { direction: "next" | "prev"; query: string }[] = [];
+  constructor() { FakeSearchAddon.instances.push(this); }
+  findNext(query: string) { this.searches.push({ direction: "next", query }); return false; }
+  findPrevious(query: string) { this.searches.push({ direction: "prev", query }); return false; }
+  clearDecorations() {}
+  onDidChangeResults() { return { dispose() {} }; }
+}
+
+/**
+ * The handler the most recently created terminal registered. useTerminal caches
+ * terminals by session id for the lifetime of the module, so a test that wants a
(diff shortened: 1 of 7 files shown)


kipavy:
Reviewed, and the diagnosis holds up — I checked the claims about `@xterm/xterm@6.0.0` against the shipped `lib/xterm.js` rather than taking the table on faith, and both are right: the ctrl+letter pass-through ends in `cancel(e, true)`, which is `preventDefault()` **and** `stopPropagation()`, so the shell path keeps the native find-next shut without help. Reverting the two source files makes 2 of the 3 new cases fail, so the regression test earns its place. The fixture extraction was the right call too.

I've pushed two commits onto the branch rather than sending you round again for something this small.

**`3c1256d5` — Ctrl+G could still move two hits per press.**

The other half of that `_keyDown` is the problem:

```js
if(this._customKeyEventHandler&&!1===this._customKeyEventHandler(e))return!1;
```

It returns early *without* calling `cancel()`. So on the path this PR keeps — widget open, handler returns `false` — the event is never stopped and goes on to `useKeyboard`'s window listener, which runs its own `next()`/`prev()` on top of the one the canvas handler just ran. Reachable by opening the widget with Ctrl+F, clicking back into the terminal (the widget only closes on Escape or the ✕), then pressing Ctrl+G. Pre-existing, not something this PR introduced, but it lives in the lines the PR rewrites and the new comment in `useKeyboard.ts` reads as though the case were handled.

The canvas handler now takes the event when it drives the widget. It is the right owner: it has the pane's own session id, where the window listener keys off `activeSessionId`. The pass-through is untouched — it has to leave the event alone so xterm can encode `^G` and cancel it on the way out.

**`ea0bdcad` — Ctrl+F had the same shape, with a worse symptom.**

`useKeyboard`'s Ctrl+F branch prefers the right panel's search bar when one is open on a searchable section. With History or Snippets open and the terminal focused, Ctrl+F opened the terminal widget and then pulled focus into the panel's search bar. The comment in `useKeyboard.ts` already asserted that `useTerminal` handles it when the canvas has focus; now that is actually true.

Both call one `claimChord()` helper instead of two copies of `preventDefault` + `stopPropagation`.

Worth writing down, because it bounds the bug: these two are the **only** branches that can double-fire. They are the two placed above `useKeyboard`'s `if (isInput) return;` guard, so they can suppress the native find dialog even inside inputs — and xterm's canvas is its helper `<textarea>`, so every other shortcut the two handlers share is already unreachable from the terminal. `matchShortcut("omni")` sits above the guard as well but is idempotent.

I renamed `useTerminal.ctrlG.test.tsx` to `useTerminal.searchChords.test.tsx`, since it now covers both chords, and added a case per chord asserting the event is claimed with the widget open and deliberately *not* claimed on the pass-through. Each fails against the code it replaces.

Local runs: `tsc --noEmit` clean, and the `useTerminal.*` / `useKeyboard.*` files pass. I could not get an honest full-suite run — my box kept OOM-killing vitest — so I am leaning on CI for that.

Issue #12, opened 2026-06-07
Termius Import Fails on Ubuntu 26.04


surbiks:
## Description

I have Termius installed on Ubuntu 26.04 using the official `.deb` package and I am logged into my account with existing hosts and data. However, importing Termius data into the application fails with the following error:

```
Termius key not found in OS keychain — is Termius installed and logged in on this machine?
```

## Investigation

I reviewed the importer's source code and checked the paths and keyring entries that it expects to find. On my system, those paths and keyring entries do not exist, despite Termius being correctly installed and actively used.

The importer appears to assume a specific storage location or keychain integration that is not present in the current Linux `.deb` version of Termius.

## Steps to Reproduce

1. Install Termius using the official `.deb` package on Ubuntu 26.04.
2. Log in to a Termius account and ensure that host data is synchronized.
3. Attempt to import Termius data into the application.
4. Observe the error:

   ```
   Termius key not found in OS keychain — is Termius installed and logged in on this machine?
   ```

## Expected Behavior

The importer should detect the installed `.deb` version of Termius and locate the required credentials or encryption keys, allowing the import to proceed.

## Actual Behavior

The importer cannot find the expected key in the OS keychain and aborts the import process, even though Termius is installed and logged in.

## Additional Information

* OS: Ubuntu 26.04
* Termius installation method: Official `.deb` package
* Termius is installed and logged in with synchronized data.
* Manual inspection suggests that the keyring locations and storage paths expected by the importer do not exist on this installation.

It would be helpful if the importer could support the current Linux `.deb` version of Termius or document any differences in how credentials and encryption keys are stored on this platform.

kipavy:
Hi, thanks for report, I'll look into it when I'll have time

kipavy:
Should be fixed in #13 — on Linux we now read the Termius key from the Secret Service (libsecret) instead of the kernel keyutils store. The v0.2.1 hotfix should be live in ~1h; the app auto-updates, so you'll just need to restart it once the release is out. Would appreciate your feedback once you can try it. 🙏

kipavy:
more like ~2h, in 1h it should be good (I had to make builds run synchronously to avoid some issues so its a lot slower than it used to be but it works.

surbiks:
thanks

PR #215, opened 2026-09-07
fix(teams): make connect-only usable, and keep VIEW_SECRETS meaningful (#190)

### LINKED ISSUE #190: connect-only members can never obtain team credentials, so the role cannot connect to anything
Found while fixing #187. Separate bug, and the one that actually makes the role unusable.

## What happens

A member holding only the built-in `connect-only` role can see every host in a team vault, and cannot connect to any of them that needs a stored password, private key, or key passphrase.

## Why

Connecting reads credentials out of the **local keychain** (`useSessionStore`, `src/stores/sessionStore.ts:421-427` — `getSecret("identity:…:password")`, `getSecret("key:…:private")`, `getSecret("key:…:passphrase")`).

For a team vault, the only thing that ever puts those entries in the local keychain is `hydrateTeamVaultSecrets` (`src/services/teamVaultSecrets.ts:77`), which needs two things a `connect-only` member has neither of:

1. `listTeamSecrets` → `GET /v1/teams/:team_id/secrets`, gated on `PERM_VIEW_SECRETS` (`server/src/routes/team_objects.rs:284`).
2. `getTeamVaultKey` → `GET /v1/teams/:team_id/vault-key`, also gated on `PERM_VIEW_SECRETS` (`server/src/routes/team_sync.rs`).

`connect-only` is `28676` in `BUILTIN_ROLES` = `CONNECT | START_TERMINAL_SESSION | JOIN_TERMINAL_SESSION | VIEW_TERMINAL_SESSIONS`. No `VIEW_SECRETS`. Both calls 403, `hydrateTeamVaultSecrets` is called under `.catch(() => {})` in `fetchTeamData`, and the failure is silent.

The host objects themselves come from `list_objects`, which is membership-gated and returns plaintext metadata — which is why the vault looks fully populated to a `connect-only` member right up until they press connect.

## Reproduction

1. Owner A creates a team vault with one SSH host that has a stored password.
2. A invites B with initial role `connect-only` only.
3. B opens the vault — the host is listed.
4. B connects — authentication fails; no password was ever stored locally. Server log: `GET .../secrets → 403`.

## The decision this needs

What does `PERM_CONNECT` mean when the credential is in the team vault?

- **`CONNECT` implies "use secrets without seeing them".** `list_secrets` and `get_my_vault_key` gate on `PERM_CONNECT || PERM_VIEW_SECRETS`; `VIEW_SECRETS` keeps meaning "reveal the plaintext in the UI" and `COPY_SECRETS` "put it on the clipboard", both of which stay denied. Honest caveat: the client decrypts in TypeScript, so this boundary is a UI promise, not a cryptographic one — a determined `connect-only` member could read the plaintext out of the running app.
- **`CONNECT` does not reach team credentials.** Then the role only works for hosts that need no stored secret (agent auth, an unencrypted key the member supplies themselves), `reconcileTeamVaultKeys` should stop wrapping vault keys for members who can never fetch them, and the UI needs to say so on the host — greying out or explaining the hosts they cannot use, rather than failing at connect time.

Either way, the silent `.catch(() => {})` should become a visible state: the member deserves to know before they press connect.

Related: #187, #70, #41.

kipavy:
Closes #190. Client half — the server half is VoltiusApp/server#12, and **this needs that deployed to actually fix the role**.

## What #190 was

A `connect-only` member sees every host in a team vault and cannot connect to any that needs a stored password, key, or passphrase. Server-side the credentials were unreachable (`list_secrets` and `vault-key` both required `VIEW_SECRETS`); server#12 widens both to `CONNECT || VIEW_SECRETS`, so the credentials now land in the member's local keychain and the role works.

## The silent failure

`hydrateTeamVaultSecrets` was called under `.catch(() => {})` in `fetchTeamData`. That is why the bug read as an authentication problem rather than a permission one: `list_objects` returns plaintext metadata, so the vault renders fully populated right up until connect fails.

It now records `credentialsUnavailableByTeamId` on the team vault state and shows a note on the hosts list in both shells. Deliberately **not** a blocked status — the vault is browsable and some hosts still connect fine — so this warns in place instead of replacing the page. A member on an older server, or one still awaiting their wrapped key, learns this before pressing connect.

## Keeping VIEW_SECRETS / COPY_SECRETS meaningful

Worth flagging, because it was not in the issue: **neither permission was enforced anywhere in the UI.** `connect-only` was contained only because the plaintext never reached the keychain. Widening the server gate without this would have promoted the role to a full member in practice — able to reveal every password in the form and export the whole vault as plaintext JSON.

The two paths that read plaintext back are now gated:

- **Editors** — `useStoredSecrets` takes the object's `vaultId` and refuses to load its secrets when the caller's role there lacks `VIEW_SECRETS`, returning `"forbidden"` so the form says why. Saving is unaffected: a field that never loaded is never dirty, and the submit path already writes `dirty ? value : null`, so nothing is overwritten with blank.
- **Export** — `ExportCtx.readSecret(vaultId)` replaces three copies of an ungated `(key) => getSecret(key).catch(() => null)` in the handlers, so a bundle cannot carry plaintext its builder is not allowed to see. Both `buildBundle` callers (ExportTab, and the plugin/MCP `exportObjects`) pass the gate; it is a required parameter, so a third caller cannot forget it.

Duplicate/move/paste flows already required `EDIT_*`, which `connect-only` does not hold, so they needed no new gate.

## Reuse

- three near-identical warning rows collapsed onto one `WarningNote`
- the selected-team lookup shared by `useBlockedTeamVault` and the new credentials hook extracted as `useSelectedTeamId`
- `resolveCan`-against-live-stores extracted as `canFromStores`, replacing the copy inlined in `plugins/runtime.ts`

## Tests

Full suite green — **3877 passed / 0 failed** across services (1085), hooks+plugins+stores (1765), components (820), i18n+lib+mcp+utils (207). `tsc --noEmit` clean.

New:
- `teamVaultSync.credentials.test.ts` — a failed hydration is recorded while the status stays `loaded`; a later success clears it
- `useStoredSecrets.test.ts` — a role without `VIEW_SECRETS` never calls `getSecret`, and the check is made against the object's own vault

The two form characterisation suites mocked `@/stores/teamStore` with `teams` alone; they now supply the roster and loader actions `usePermissions` reads.

## Not covered

Not live-verified with two real accounts yet — that needs server#12 deployed. The strings are translated in all four locales but owe a native pass (fr/ru/zh).


kipavy:
## Live-verified end-to-end, two real accounts (2026-09-07)

Two app instances built from this branch (separate `VOLTIUS_KEYCHAIN_NS`, so genuinely independent accounts) against a throwaway server built from VoltiusApp/server#12 and a throwaway Postgres. Prod and the shared `tauri-headless` were never touched.

Setup, all through the real UI: **A** registers → creates vault `Team190` → turns it into a team vault → adds host `ssh-host-1` with a stored password → invites **B** with `connect-only` as the only initial role → B accepts. Server state confirms the fixture is the real #190 shape:

```
proud-beacon-7412 | connect-only | 28676     <- B, no VIEW_SECRETS
rapid-beacon-0311 | owner        | 131071
team_vault_keys: both members hold a wrapped key
team_vault_secrets: password:<conn-id> | connection_password | ciphertext
```

### 1. The bug is fixed — B connects

Server log for B (`user_id=b1d2fec7…`):

```
GET  /vault-key  -> 200   Vault key fetched user_id=b1d2fec7…
GET  /secrets    -> 200
PUT  /secrets    -> 403   (writes still denied — no EDIT bits)
```

and on `ssh-host-1`:

```
sshd-session.pam: voltius [priv]
sshd-session.pam: voltius@pts/0
tcp 172.20.0.2:2222  172.20.0.8:56470  ESTABLISHED
```

`172.20.0.8` is B's container. A never connected in this run. So a `connect-only` member authenticated with a password they could only have obtained through `/vault-key` + `/secrets` — the two routes that returned 403 before the server fix. The session stayed up for minutes. B's hosts list showed **no** credentials warning, which is the correct outcome when hydration succeeds.

### 2. The VIEW_SECRETS gate holds for the same member

Granting B `EDIT_CONNECTIONS` **without** `VIEW_SECRETS` (`28684`) makes the host editor reachable. Opening it:

```
pwFields=2 values=["",""]  forbiddenNote=true
"Your role in this vault cannot see stored secrets, so these fields are left blank.
 Connecting still works — the saved credential is used without being shown."
```

The password is in that member's local keychain — they had just connected with it — and the editor still refuses to render it. This is the property that would have been lost if the server gate were widened alone.

With plain `connect-only` (no edit bits) the context menu has no **Edit** row at all, so the gate is defence in depth behind the existing `EDIT_*` gating.

### 3. The silent `.catch(() => {})` is gone

Stripping both `CONNECT` and `VIEW_SECRETS` from the role (keeping `EDIT_CONNECTIONS` + session bits) reproduces the old failure mode on the wire:

```
GET /objects    -> 200
GET /vault-key  -> 403
GET /secrets    -> 403
```

and the vault now says so instead of looking healthy:

> Stored credentials for this team vault could not be loaded, so hosts that need a saved password, key, or passphrase will fail to connect. Ask a vault administrator to grant your role access to the vault's secrets.

`cards=1` at the same moment — the host list still renders, confirming this is a warning in place and not a blocked status.

Issue #219, opened 2026-09-07
Seat pre-check uses purchased seats, not the trial-clamped cap — invite still 402s


kipavy:
Found while auditing #72. Self-contained server fix; does not depend on any of that issue's UI work.

### What happens

`/subscription-info` reports the **raw purchased** seat count:

```rust
let seats = if (tier == "teams" || tier == "business") && row.2.is_none() {
    Some(3)
} else {
    row.2          // users.seat_count, unclamped
};
```
`src/routes/billing.rs:518-523`

But the invite path enforces a **trial-clamped** cap:

```rust
/// The owner's effective seat cap, or `None` when uncapped. An active trial
/// clamps the cap to 10 however many seats were purchased.
Ok(seat_count.map(|seats| {
    let effective = if trial_ends_at.is_some() { seats.min(10) } else { seats };
    effective as i64
}))
```
`src/routes/teams.rs:47-62`

The client reads `seats` into `totalSeats` (`src/stores/subscriptionStore.ts:104`) and pre-checks against it before sending an invite (`seatAvailability` → `isAtSeatLimit`, `src/components/members/MembersPage.tsx:682`, `:720`).

**Result for a trial owner who purchased 25 seats and has 4 in use:** `SeatsMeter` reads "4 of 25 used, 21 available", `isAtSeatLimit` is false, the client sends the invite, and the server rejects with 402 at 10. The seat wall reappears in exactly the case nothing explains — and the pre-check that exists specifically to avoid the transactional 402 is silently bypassed.

### Second problem: the meter's scope is wrong

`used_seats` counts distinct users across **every team the owner owns**:

```sql
SELECT COUNT(DISTINCT tm.user_id)
FROM team_members tm
JOIN teams t ON tm.team_id = t.id
WHERE t.owner_id = $1
```
`src/routes/billing.rs:501-506`, matching `owner_seats_used` at `teams.rs:65-75`.

`SeatsMeter` renders that number inside a single team's invite panel with no indication it is account-wide, so an owner with several teams sees a number that does not match the members in front of them.

### Third: `add_member` reimplements the clamp

`invite_member` uses the helpers (`teams.rs:1238-1239`), but `add_member` inlines its own copy of both the clamp and the used-count query (`teams.rs:404-429`) instead of calling `owner_seat_cap` / `owner_seats_used`. Three copies of the same rule, one of which is the one leaking to the client. Worth collapsing as part of this fix rather than adding a fourth.

### Fix

[...]

Issue #217, opened 2026-09-07
Rotate the team DEK on membership change (defense in depth, not offboarding)


kipavy:
Split out of #72 so its rationale is recorded honestly rather than assumed.

### Current behaviour

Removing a member deletes their row from `team_vault_keys` (`server/src/routes/teams.rs:545-550`) and revokes their session grants, and the client evicts the in-memory vault key (`src/services/sync.ts:882-886`). The team DEK itself is never rotated, and the team blob is never re-encrypted.

### What rotation would and would not buy

It is tempting to treat rotation as the fix for offboarding. It isn't, because a departing member held the plaintext:

| Exposure | Protected today? | Does rotation help? |
|---|---|---|
| Secrets they already read | No — permanently compromised | No. Only rotating the credentials themselves helps |
| Future team content | Yes — membership row, wrapped key row and session grants all deleted server-side | No, absent a server-authz bypass |
| Blob ciphertext they exfiltrated before leaving | They held the DEK, so they can decrypt it | No — a copy someone already holds cannot be retroactively re-encrypted |
| Plaintext left in their local keychain | No — see #216 | No |

Rotation protects only blobs written **after** rotation, which a removed member can no longer fetch anyway.

### The case for doing it anyway

Defense in depth on the ciphertext at rest. Today a single DEK covers the entire lifetime of a team, so any leak of the encrypted blob — a database backup, a stale replica, a misconfigured storage bucket — exposes everything the team has ever stored, including data added long after a member left. Rotating on membership change bounds that blast radius to one membership epoch.

That is a real property, and it is worth stating as the reason. It is not an answer to "what happens to a removed member's access", which is #72's question.

### Rough shape

- Generate a new team DEK, re-encrypt the team blob, re-wrap for every remaining member.
- Handle partial rotation: some members re-wrapped, some not, and the blob in one state or the other. This needs to be atomic or resumable — a half-rotated team where some members cannot decrypt is worse than no rotation.
- Handle offline members: someone who is not connected at rotation time must re-key on next connect without losing access in the meantime.
- Decide the trigger: automatic on every removal, or an explicit admin action. Automatic is safer but makes every removal a heavyweight operation.
- Decide what "rotate" does to older blob versions retained server-side, if any.

### Related

[...]

Issue #72, opened 2026-07-24
Team administration & offboarding — roles, pending-invite status, seats, leave/remove


kipavy:
Part of the **team-ux** initiative (siblings: #65–#69).

**Relates:** #67 (pending-invite management overlaps the inbox), #68 (roles & unified panels), #71 (hub is the likely home for admin entry).

### Problem
The inviter/admin surface (`MembersPage.tsx`) is dense and under-designed, and the offboarding side is essentially absent.

**Administration (inviter side):**
- **Roles** — assigning/changing roles (owner/manager/editor/member/connect-only) is present but not clearly explained; no per-role "what can this person do" surface.
- **Pending invites** — status is thin: revoke is a hover-`x` on `PendingInviteCard`; there's no clear *resend*, no "invited but not yet accepted" clarity, no expiry.
- **Seats / paywall** — the seats usage bar → HTTP 402 → `BuySeatsModal` path works but is transactional and easy to hit as a wall; the "why is my invite blocked" moment isn't well handled.
- **Two-step member removal + undo** exists but isn't a coherent, reassuring flow.

**Offboarding (missing):**
- A member **leaving** a team — no clear action or consequence explanation.
- An owner **removing** a member — what visibly happens to that person's access and keys (revocation, re-wrap) is neither surfaced to the admin nor to the removed user.

### Proposed reflow
- **Clear role model:** show what each role grants at the point of assignment; make changes obviously reversible.
- **First-class pending-invite management:** status (invited / accepted / expired), **resend** and **revoke** as explicit actions, consistent with the receive-side inbox (#67).
- **Seat UX that isn't a wall:** surface seat usage *before* the invite is blocked; make the buy-more path calm and explain what a seat is.
- **Explicit offboarding:** "Leave team" and "Remove member" as honest flows that state the access/key consequences, with confirmation.

### Decisions (brainstorm 2026-07-24)
- Filed as one issue (per brainstorm) since it's all the same admin surface; **may split** into "member/role/seat administration" and "offboarding" if it grows.
- Role set and panel unification are coordinated with #68 (don't diverge role vocabularies).

### Still open
- [ ] Whether admin lives in the Team hub (#71), stays per-vault in `MembersPage`, or both.
- [ ] Key-revocation UX + what the removed user sees.
- [ ] Seat model presentation (per-user? explain trial seat-clamp behavior).
- [ ] Split vs. keep-as-one decision.

kipavy:
[...]

PR #225, opened 2026-09-08
feat(teams): offboarding, invite lifecycle, and the keychain leak on removal (#72, closes #216)

### LINKED ISSUE #216: Team secrets survive removal in the departed member's OS keychain
Found while auditing #72. Independent of every UX item in that issue, and small to fix.

### What happens

`fetchTeamData` writes every team secret into the member's local OS keychain — connection passwords, private keys, passphrases:

```ts
for (const [k, v] of Object.entries(blobPayload.secrets ?? {})) {
  await storeSecret(k, v).catch(() => {});
}
```
`src/services/teamVaultSync.ts:388`

The only function that removes them from disk is `_clearTeamStores` (`teamVaultSync.ts:499-524`), which issues `deleteSecret` for each `password:<id>` / `key:<id>` / `passphrase:<id>` / `identity:<id>:password`. It is called **exclusively** from `fetchTeamData`'s own error and 403 branches (`teamVaultSync.ts:288`, `:316`, `:341`, `:352`, `:358`, `:367`).

The removal path does not call it. `onTeamRemoved` clears the seven team store slices individually (`src/services/sync.ts:918-925`), and those are pure in-memory Zustand map deletes (`src/stores/connectionStore.ts:75`).

**So when a member is removed from a team, the team's plaintext passwords and private keys stay in their OS keychain indefinitely.**

### Why it can't self-heal later

`_clearTeamStores` needs the object IDs to build the `deleteSecret` keys, and the code says so:

> `// Wipe secrets from disk before clearing in-memory state so we still have the IDs.`

`onTeamRemoved` clears memory first. Even if a later `fetchTeamData` hit the 403 branch, `_clearTeamStores` would read empty arrays from the stores and delete nothing. The ordering is what makes it permanent.

### Scope

Everything else about removal is already correct — the server deletes the `team_vault_keys` row, revokes the departed member's session grants and pushes `membership_changed` (`server/src/routes/teams.rs:545-563`); the client evicts the in-memory vault key and unlinks the local vault. Only the on-disk secrets are missed.

Note this is not addressed by rotating the team DEK: the departing member held the plaintext, and these entries *are* plaintext in their keychain. Rotation is a separate concern.

### Fix

Export `_clearTeamStores` as `clearTeamStoresAndSecrets` and call it from `onTeamRemoved` in place of the seven individual `clearTeamXxx` calls. One substitution: it closes the leak and removes seven lines of duplicated clearing the service already maintains.

Ordering is the whole fix, so the regression test must assert `deleteSecret` fired for the team's connection, key and identity secret names **before** the stores went empty — a test that only checks the secrets are gone at the end would pass against the broken code if the calls were merely reordered wrong.

### Related

- #72 — team administration & offboarding; this is gap 2 of that audit, pulled out because it shouldn't wait behind a UI decomposition.
- Design context: `docs/superpowers/specs/2026-09-07-team-admin-offboarding-design.md`

kipavy:
Implements the offboarding and invite-lifecycle half of #72, and closes #216.

An audit of #72 against the current code found most of what it asks for already shipped, and two real bugs underneath it that the issue does not mention. The audit is [posted on #72](https://github.com/VoltiusApp/voltius/issues/72#issuecomment-5575542719); the design is in `docs/superpowers/specs/2026-09-07-team-admin-offboarding-design.md`.

### The security fix (#216)

`fetchTeamData` writes every team secret into the member's local OS keychain — connection passwords, private keys, passphrases (`teamVaultSync.ts`). The only function that removed them from disk was called exclusively from `fetchTeamData`'s own error branches, never from the removal path, which cleared the in-memory Zustand slices and nothing else.

**A removed member kept the team's plaintext passwords and private keys in their OS keychain, permanently.**

The ordering is the whole fix: the wipe builds its `deleteSecret` keys from the object IDs, so it must run *before* the stores are emptied — the existing code says so in a comment, and `onTeamRemoved` was clearing memory first. `clearTeamStoresAndSecrets` is now called from the removal path, and a regression test asserts an already-emptied store deletes nothing.

### What else changed

- **The removed user is told.** New `membershipEnded` inbox kind plus a toast, naming the team and what to rotate. `InboxKind`'s contract says every kind is actionable, so its action opens the keychain page. Suppressed when you left voluntarily.
- **"Leave team".** The server already allowed self-removal (`remove_member` only requires `MANAGE_MEMBERS` when `auth.0 != user_id`); this was missing client UI, not a missing endpoint. Not offered to an owner, whom the server rejects.
- **One removal path.** Removal was written **four** times, each with its own snapshot and undo — the detail panel (two-step confirm), the context-menu kick (no confirm), the bulk kick (no confirm), and one in Settings › Vaults (two-step confirm, *no undo at all*). All now route through `teamOffboarding.departMembers` behind a shared confirmation dialog, so **bulk kick gains the confirmation it never had** as a consequence rather than as separate work.
- **Honest consequence copy.** The dialog states what a departure does and does not undo: access ends everywhere, the vault leaves their devices, and they keep whatever they already saw — rotate those credentials.
- **Invite lifecycle.** Expired invitations used to vanish from the admin's list silently. They now stay, labelled, with sent-date, an expiry countdown, and **Resend** and **Revoke** as real buttons rather than a hover-only `×`. Resend reuses the existing invite path, which already upserts and pushes the expiry out.
- **Roles explained where they are granted.** `members.roleBlurb.*` already existed and was rendered only by `InviteControl`; it now also appears under the chips in the invite panel and the member detail panel, alongside the permission tooltip.
- **`MembersPage.tsx` split**, 1629 → 691 lines, into `MembersToolbar`, `roleBadges`, `cards/`, and `panels/`. The private-vault self card was written twice inline (grid and list differing only by layout) and is now one component.

Needs the server side for the invite status: https://github.com/VoltiusApp/server/pull/13.

### Verification

- `pnpm vitest run src/components src/services src/stores` — green
- `tsc --noEmit` — exit 0
- **Look-preservation pixel diff for the split.** One navigation script, run before and after, PIL diff in a headless container. It caught a real regression that tests and types both passed: `SelfCard` built its badge class as a template appending `px-1` after `px-1.5`, leaving both on the element — Tailwind resolves that by stylesheet order, not string order, so the grid badge silently took the wrong padding. A control pass (identical code captured twice) established that an ~18×16 region top-right differs on its own; without it I'd have either chased that noise or waved through the real 1071-pixel difference. Final result: **zero differing pixels outside that proven-noisy region**, in both list and grid.

### Not covered

Every test here mocks `@/services/vault`, so **none of them prove the real OS keychain is empty after a removal** — which is exactly why #216 stayed invisible. That needs a live two-account run before this is trusted in a release.

Split out of #72 and not in this PR: #217 (team DEK rotation — deliberately scoped out; it does not address what a departing member already read) and #219 (the trial seat-clamp mismatch, a server-only fix).


kipavy:
### Live two-account run — done, and it changed the PR

The "Not covered" caveat in the description is now resolved. Ran the full flow on an isolated stack (own server + Postgres + two app instances with separate keychain namespaces; prod untouched and verified before/after).

**#216 confirmed fixed**, using a temporary dev-only probe that exposed the app's own `getSecret` — file size/mtime on `secrets.enc` turned out to be useless as evidence, because a control run against the pre-fix code showed the file is rewritten at removal *anyway*. The probe was removed before committing.

```
bob in team   -> getSecret(password:06d02ec1…) = PRESENT:SUPERSECRET-E2E-PASSWORD
bob removed   -> getSecret(password:06d02ec1…) = NULL
bob leaves    -> PRESENT -> NULL, and no removal notice (voluntary suppression works)
```

Also confirmed live: the offboarding dialog renders all four consequence points; **Leave team** appears on your own card and not on the owner's; the `membershipEnded` inbox entry appears with a working **Review credentials** action that opens the keychain page; the role blurb renders in both the invite panel and the member detail panel; and the collapsed `SelfCard` renders correctly.

### One real bug found and fixed (1c4ff2d4)

The removal notice read **"You were removed from a5d25b38-f904-4f52-b03b-e629445dbf3b"** — the raw team id. `handleMembershipChangedEvent` calls `loadTeams()` before `onTeamRemoved`, so by the time the handler runs the team is already out of the store and my `?? tid` fallback fired. Unit tests passed the name in directly, so nothing could have caught this short of running it. Fixed by snapshotting id→name before the diff; `teamMembershipEvents.test.ts` now pins that ordering so a future refactor cannot silently reintroduce it.

### Wording correction

The description says "OS keychain". Secrets actually live in the app's encrypted `secrets.enc` store, with the OS keychain holding the key that opens it. Same impact, more accurate mechanism.

PR #230, opened 2026-09-08
fix(accounts): carry wrapped_user_secrets per account (#228)

### LINKED ISSUE #228: Switching accounts breaks the team vault key: wrapped_user_secrets is a global keychain slot, not per-account
## Symptom

Switch back to an account through the quick switcher and its **team vaults stop opening** — `Couldn't load team vault · Something went wrong. Try again`. Try again re-runs the identical failure. A webview reload does not help. Signing out completely and signing back in **with the master password does** fix it.

Found while live-testing #68 with two cloud accounts on an isolated server.

## Repro

1. Register cloud account **A**. Convert a private vault to a team vault (so A holds a wrapped team vault key).
2. Account menu → **Add another account…** → register cloud account **B**.
3. Account menu → switch back to **A**.
4. Open A's team vault.

Observed at step 4:

- `GET /v1/teams/<id>/vault-key` returns **200** — the wrapped key is served fine.
- The DB is consistent: A's row in `team_vault_keys` is `wrapped_by = A`, and A's `users.public_key` is the key it was wrapped to.
- The client still fails, so the break is local: the wrapped key cannot be unwrapped.

## Mechanism

`wrapped_user_secrets` is a **single, un-namespaced keychain slot**, and the switcher does not carry it per account.

1. `createServerAccount` (`src/services/account.ts`) generates a random **DEK** + an x25519 private key, wraps both under the **KEK** (`deriveKeys(password, account_id)`), and writes `keychainSet("wrapped_user_secrets", …)`. The vault key becomes the DEK. Registering **B overwrites A's value**.
2. `SESSION_KEYS` (`src/services/savedAccounts.ts:18`) is `account_id, mode, master_password, email, server_url, jwt, refresh_token`. **`wrapped_user_secrets` is absent**, so `switchToAccount` restores everything except the thing needed to recover the DEK.
3. `switchToAccount` reloads, and the reload runs `autoLogin` — documented *"local Tauri calls only — autoLogin stays instant offline"*. It calls `passwordVaultKey(kek)` → `adoptUserSecrets(A_kek, <B's wrapped secrets>)` → unwrap fails → falls back to `keyThatOpensVault(kek)`.
4. `multiplayerService.getMyX25519Keypair()` derives the keypair from `getVaultKey()`. A different vault key means a different x25519 private key, so `unwrapSessionKey(wrapped_key, wrapper.public_key)` fails, `getTeamVaultKey` throws `"error"`, and the panel shows the generic failure.

The healing asymmetry is the confirming evidence: `login()`'s `reauth` branch (`account.ts:320-328`) re-fetches `wrapped_user_secrets` **from the server**, unwraps it with the correct KEK, and rewrites the keychain slot. `autoLogin` never contacts the server. That is exactly why a reload does nothing and a full password sign-in heals it.

## Secondary problems this exposes

- **The error is unactionable.** `Something went wrong. Try again` re-runs the same failing unwrap. Nothing hints that signing out and back in is the fix. A vault that cannot be unwrapped locally is a different state from a transient load error and deserves its own copy.
- **Two sources for one x25519 identity.** `useVaultKeysStore` holds an `x25519Private` recovered from the wrapped secrets, but `getMyX25519Keypair` ignores it and re-derives from the vault key. `autoLogin` populates neither, so after a switch the store is empty and only the derived one is in play.
- ⚠️ **Possible self-worsening — not observed, flagging for review.** `sync.ts`'s `completeTeamLoginSetup` calls `getMyX25519Keypair()` then `updatePublicKey(publicKey)` unconditionally (best-effort, errors swallowed). In the broken state that would publish a public key derived from the *wrong* vault key. Since `getTeamVaultKey` resolves the wrapper's key from the roster, the roster would then advertise a wrong key for that user. In my run A's `users.public_key` was still correct and a fresh sign-in healed everything, so I did **not** see this fire — but the call is unguarded, and it is worth confirming it cannot overwrite a good published key with a derived-from-wrong-DEK one.

## Suggested fix

Namespace `wrapped_user_secrets` per account (or add it to `SESSION_KEYS` so `saveCurrentAccount` / `switchToAccount` carry it like every other per-account value). Either way `autoLogin` then adopts the right secrets and derives the right identity offline.

Worth pairing with: refuse to publish a public key when the DEK was not successfully adopted, rather than pushing a derived-from-fallback key.

## Confidence

The observations are direct: 200 from the server, consistent DB rows, symptom only after a switch, reload no help, fresh sign-in heals, and `Grant now` working immediately afterwards. The mechanism above is derived from reading the code, **not** instrumented — a log of the derived key bytes before and after a switch would confirm it. The isolated test containers have since been removed.

https://claude.ai/code/session_014gY5k7XbeR5zoTXdzsuWxF

kipavy:
Closes #228.

## The bug

`wrapped_user_secrets` is a per-account keychain slot that was missing from
`SESSION_KEYS`, so `tearDownSession` deleted it and `switchToAccount` restored
everything except the value that recovers the dek. `autoLogin` then found no
cached secrets and fell back to `keyThatOpensVault(kek)`, which accepts the kek
unconditionally against the freshly wiped vault. The x25519 identity is derived
from the vault key, so the account came back on a different identity and the
team vault key wrapped to the real one no longer unwrapped.

## Changes

**The slot is a session key now.** `saveCurrentAccount` snapshots it,
`switchToAccount` writes it back. `upsertSavedAccount` also stops treating null
as a supplied value, so a snapshot taken mid-teardown cannot blank the stored
one.

**autoLogin heals an account that has no cached secrets.** For `mode ===
"server"` it asks `/auth/me` — the same fetch that makes a password sign-in
recover — with a 5s budget, and only when the slot is empty. That is what
repairs switcher entries saved before this PR. A server answer with no
`wrapped_user_secrets` means a pre-split account, whose kek-encrypted vault and
kek-derived identity are already correct.

**A session that could not prove its vault key no longer publishes an identity.**
When the secrets can be neither read nor fetched the session is marked
`identityUnproven` and `publishMyPublicKey` refuses. Unguarded — as
`completeTeamLoginSetup` was, flagged in the issue as unobserved — it would push
a key derived from the wrong vault key over a good roster key, after which every
teammate wraps to a key nobody holds. That derive-then-publish shape existed four
times (`sync`, `teamVaultSync`, twice in `multiplayerService`) and is now one
function.

**A local unwrap failure has its own state.** `getTeamVaultKey` threw the generic
`"error"`, whose panel offers Try again, which repeats the identical failing
unwrap. It now throws `"key_mismatch"` and the panel names the sign-in that
recovers the secrets instead of offering the button.

## Found on the way: registration published the wrong public key

`createServerAccount` registered `generate_user_secrets_cmd`'s random
`x25519_public`, while every wrap and unwrap uses
`HKDF(vault key, "x25519_keypair_v1")`. Until the first `updatePublicKey`
landed, the roster advertised a key whose private half nothing on the device
reaches for — a teammate wrapping in that window produced exactly the
`key_mismatch` this PR now names. Both registration paths publish the identity
their session's vault key derives: the dek for a new cloud account, the kek for
a local vault linking to a server, which linking does not rekey. The two copies
of the register call are now one.

The stored `x25519Private` remains vestigial — it opens nothing, and every
published roster key is the derived one — so `getMyX25519Keypair` still ignores
it, now with a comment saying why.

## Verification

`tsc --noEmit` clean; the full frontend suite passes in chunks (3991 tests
across `src/services`, `src/stores`, `src/hooks`, `src/components`, `src/i18n`,
`src/utils`, `src/lib`, `src/app`, `src/commands`, `src/plugins`, `src/mcp`,
`src/themes`). New tests cover the switch carrying the slot, the null-merge
guard, all four autoLogin recovery outcomes, `key_mismatch`, its panel copy, and
the registered public key for both registration paths.

Not live-verified: the issue's isolated two-account containers were removed, so
this is unit-tested and read-derived, not driven against a running server.

https://claude.ai/code/session_014gY5k7XbeR5zoTXdzsuWxF


kipavy:
## Live-verified with two cloud accounts

Isolated stack: `voltius-e2e-server` + `postgres:16-alpine` on a private docker network, a private `tauri-mcp` container serving this branch, its own keychain namespace. Prod untouched.

**Setup.** Registered `a@e2e.test`, converted its Personal vault to a team vault (A holds a wrapped team vault key), added a host, then registered `b@e2e.test` through *Add another account…*, then switched back to A.

### 1. The slot is carried per account

```
SESSION_KEYS = [account_id, mode, master_password, email, server_url, jwt, refresh_token, wrapped_user_secrets]
saved: [ {a@e2e.test, hasWrapped: true, "AU0JiJ6gP95O…"},
         {b@e2e.test, hasWrapped: true, "AaLAdW2fqY9D…"} ]
```

Each prefix matches that user's `users.wrapped_user_secrets` row. B's registration overwrites the live slot, which is what used to strand A.

### 2. After switching back to A

| | before the switch | after |
|---|---|---|
| `dek` / `x25519Private` in store | present | present |
| derived public key | `/Vj1huLBfpWbOd…` | `/Vj1huLBfpWbOd…` |
| `getTeamVaultKey` | unwrapped, 32 bytes | unwrapped, 32 bytes |
| vault status | `loaded` | `loaded` |

### 3. Control: the same run on `origin/dev`'s `account.ts` + `savedAccounts.ts`

Reverted just those two files in the served worktree, repeated the switch:

```
store:  { hasDek: false, hasX25519: false }
derived public key: oHa49L6ZMTDk83LJ…      ← the kek-derived one, not A's
getTeamVaultKey: THREW                     ← "error" on dev; this branch labels it key_mismatch
vault status: key_mismatch
```

So the failure is the one #228 describes, and the fix is what removes it.

### 4. The registration mismatch, confirmed

`users.public_key` for A is `/Vj1huLBfpWbOdMya556Dpm0iIAIdSTd8GJUQiAjigM=`, byte-identical to what `getMyX25519Keypair()` derives in the running app. Before this branch it was `generate_user_secrets_cmd`'s random key.

### 5. The escalation the issue flagged as unobserved — it fires

During the control run the broken session **published its wrong key**: A's `users.public_key` became `oHa49L6ZMTDk…`. Since `getTeamVaultKey` resolves the wrapper from the roster, every later wrap would have gone to a key nobody holds.

`identityUnproven` did not catch it. That flag is set when the wrapped secrets can be neither read nor fetched; here they *were* recovered from `/auth/me`, and the session still ran on the kek because the stale `secrets.enc` the broken session wrote only opens with the kek.

Fixed in d9ec188: an account's identity never changes, so `publishMyPublicKey` now refuses to overwrite a published key it does not derive, and marks the session unproven. Re-run with A's roster key restored:

```
identityUnproven: true
users.public_key: /Vj1huLBfpWbOd…   ← unchanged; the previous run had overwritten it
```

### 6. The recovery the new copy promises

`key_mismatch` says *"Sign out of this account and sign back in with your master password."* Did exactly that: identity back to `/Vj1huLBfpWbOd…`, vault status `loaded`, key unwrapped, roster intact.

### Checks

`tsc --noEmit` clean. `src/services` green (123 files / 1117 tests).

https://claude.ai/code/session_014gY5k7XbeR5zoTXdzsuWxF

Issue #15, opened 2026-06-15
Feature request: Import from ZOC Terminal


kipavy:
https://www.emtec.com/zoc/index.html

PR #237, opened 2026-09-09
feat(teams): encrypt team vault object metadata (#229)

### LINKED ISSUE #229: team_vault_objects.metadata is stored in plaintext server-side
Found while triaging #217. The team DEK protects the credential column, but the objects those credentials belong to are stored server-side in the clear.

### What is stored in plaintext

`migrations/021_team_vault_objects.sql` splits team vault data into two tables:

```sql
CREATE TABLE team_vault_objects (
    ...
    name        TEXT,
    metadata    JSONB NOT NULL DEFAULT '{}'::jsonb,
    ...
);

CREATE TABLE team_vault_secrets (
    ...
    ciphertext  TEXT NOT NULL,
    ...
);
```

Only `team_vault_secrets.ciphertext` is encrypted under the team DEK. `team_vault_objects.metadata` is raw JSONB, and it holds the entire object.

`saveTeamVaultObject` passes the object through untouched:

```ts
await upsertTeamObject(teamId, {
  object_id: item.id,
  object_type: objectType,
  name: item.name,
  folder_id: item.folder_id,
  metadata: item,
});
```
`src/services/teamObjectPersistence.ts:12-20`

For a connection, `item` is built at `src/stores/connectionStore.ts:81-118` and carries `host`, `port`, `username`, `auth_type`, `tags`, `identity_id`, `key_id`, `jump_hosts`, `env_vars`, `pre_command`, `post_command`, `pre_snippet_id`, `post_snippet_id`, `terminal_encoding`, `distro` and `notes`.

Nothing removes any of it on the way out — `upsertTeamObject` serialises the object as-is (`src/services/teamObjects.ts:97-104`) — and the server binds it directly into the column (`server/src/routes/team_objects.rs:202`).

### Why it matters

Anyone with read access to the database, a backup, a replica or a snapshot can reconstruct the team's full infrastructure map without holding any key: which hosts exist, who logs into them, which jump hosts front them, and what commands run on connect. `env_vars` and `pre_command`/`post_command` are free-text fields that users do in practice put tokens into, so this is not strictly limited to non-secret data either.

This is the same threat model #217 argues from — leaked ciphertext at rest — but it covers more of the sensitive surface, and unlike #217 it is exploitable with no key material at all.

### Scope

Two viable designs, and the choice is the main open question:

- **Encrypt `metadata` under the team DEK.** Strongest. Costs the server every query that filters or sorts on those fields — `idx_team_vault_objects_team_type` and the `name` column exist because the server does read them today, so this needs an audit of what actually depends on server-side visibility.
- **Strip the sensitive fields into an encrypted sidecar**, keeping only what the server genuinely needs (`object_id`, `object_type`, `vault_id`, `folder_id`, timestamps) in the clear. Cheaper, keeps sorting and filtering, but leaves the object graph shape visible and requires deciding field by field.

Either way it needs a migration path for rows already written in the clear, and re-encryption on the client for every existing team.

### Related

- #217 — rotate the team DEK on membership change. This should land first; see https://github.com/VoltiusApp/voltius/issues/217#issuecomment-5585612513 for why rotation is misleading as defense in depth while this is outstanding.
- #216 — team secrets left in a departed member's OS keychain. Independent, and the highest priority of the three.
- #72 — team administration and offboarding.

kipavy:
Closes #229. Unblocks #217.

Client half. The server half is VoltiusApp/server#15 and the two have a required deployment order, at the bottom.

## The problem

`saveTeamVaultObject` sent the whole object as `metadata: item`, nothing stripped it, and the server stored it as raw JSONB. So `host`, `port`, `username`, `jump_hosts`, `env_vars`, `pre_command`, `post_command` and `notes` were readable from any database, backup or replica **with no key at all**. Only `team_vault_secrets.ciphertext` was ever encrypted.

## What this does

Each object's metadata is now encrypted client-side under the team DEK, in an envelope the server treats as opaque. No new crypto — it reuses the `encrypt_payload` / `backup_decrypt` Tauri commands that team secrets already use.

| Commit | |
|---|---|
| `dd12fe50` `058d2058` | The `{ v: 2, enc }` envelope. `isEncryptedEnvelope` requires **both** `v === 2` and a string `enc`, so a legacy object that happens to carry a `v` field is not misread. Decode throws on every failure — bad base64, failed auth tag, and a payload that decrypts but carries no metadata key. |
| `0d7e12e8` | Read path. Decodes legacy plaintext and v2 rows in the same team, so a team mid-migration stays readable. A row that will not decrypt is **dropped**, not spread — spreading yields `{ v, enc, updated_by }` with no `id` or `host`, which renders as a broken entry and gets written back on the next save. |
| `a61dfc9b` `c62a44cc` | `getTeamVaultKey` fixes forced by the above. Decoding per row meant one key fetch per row; it cached only the resolved value, so a cold cache produced N concurrent fetches against a rate-limited route, and a 429 became a dropped row. Now memoizes the in-flight promise. Separately, an in-flight fetch could resolve *after* `deleteTeamKey` and write the key back for a team the user had just been removed from — a per-team eviction generation now rejects that, checked inside the shared promise so concurrent joiners are covered too. |
| `5e7269e2` | Write path. Sends the envelope, with `name`/`folder_id` as null. |
| `03a428a0` | One-time re-encryption pass for rows written before this. Permission-filtered client-side (the server rejects a whole batch containing a type you cannot edit), batched, idempotent, resumable, and deliberately not routed through `saveTeamVaultObject` — that stamps the audit log, and re-encryption is a migration, not an edit. |
| `bf2dbe9b` `981f7be0` | `X-Client-Version`, and a `426` surfaced as a classifiable error. The fix commit matters: the first version cached a `"0.0.0"` fallback permanently, so one transient `getVersion()` failure would have marked a current build as below-floor for the whole session. It now caches the in-flight promise, clears it on rejection, and omits the header when the version is unknown rather than sending a fabricated one. |
| `5d452871` | Settings shows how many objects a team still has unencrypted, with the reason and the resolution. |
| `ad42ab93` | The strings in Russian and Chinese — the suite ships four locales and `keyParity.test.ts` enforces it. |
| `97943d90` | Whole-branch review follow-ups, below. |

## Review follow-ups in `97943d90`

- **Deleted `saveTeamData`.** It had no callers. It was the legacy collect-stores-encrypt-upload path — exactly the shape that would turn a row this client failed to decrypt into a deletion for the whole team. Leaving it exported invited a future caller to reintroduce the hazard the read path is careful to avoid.
- **The re-encryption pass now joins an in-flight pass** instead of duplicating it. `fetchTeamData` serialises per team, but the pass is fire-and-forget and escaped that queue, so two loads in quick succession both encrypted and PUT the same rows.
- **Extracted `isStillPlaintext`** — the "alive and still in the clear" guard was written twice in inverted forms.

## Known and accepted

- A row whose metadata is null/empty is kept rather than dropped. Pre-existing, and the column is `NOT NULL DEFAULT '{}'` server-side.
- `_teamKeyGeneration` never deletes entries — one integer per team, process lifetime.
- `initTeamVaultKey` writes the key cache directly, bypassing the generation guard. A separate, deliberate provisioning path.
- A team with zero objects never runs the pass, so no count is recorded; absent renders as nothing, which is intended.
- Object counts, types, timestamps and acting user ids remain visible server-side. Inherent to routing these rows without a key.

## What this does not fix

A team whose only edit-privileged members never reconnect keeps plaintext metadata indefinitely. The server cannot fix it — it holds no key. That is what the settings count surfaces rather than hides.

## Verification

- 512 test files, 4009 tests, 0 failed. Run in chunks; a full `vitest run` is OOM-killed on the dev box.
- `tsc --noEmit` exit 0.

## Deployment order

VoltiusApp/server#15 first with the version floor **unset**, then this release, and only then set the floor. Setting it before this ships locks every client out of team vault writes.


kipavy:
GitGuardian false positive triage: all 3 findings (37087843/37087844/37087845) are in `sync.teamRemovedFailures.test.ts`, asserting keychain key identifiers like `password:c-t1` / `password:c-t2` get wiped — not credential values. This commit is a merge from `dev`, predates this PR's own changes. No dashboard access to dismiss on the GitGuardian side; merging past the red check on that basis.

Issue #244, opened 2026-09-09
Two duplicated shapes left by the vault-admin rework


kipavy:
Follow-up from #222 / #239. Both are cases the repo's own DRY gate covers ("Before writing a second caller of a shape that already exists — extract the shape and call it twice").

## 1. The menu + dialogs mount pair, duplicated across both hosts

`VaultHeader` and `VaultSidebar` each end with the identical two-line pair, differing only in `onClose`:

```tsx
{admin.pos && <ContextMenu items={admin.items} pos={admin.pos} onClose={admin.closeMenu} />}
{target && <VaultAdminDialogs target={target} dialog={admin.dialog} onClose={…} />}
```

A third caller already exists — `VaultsSection` renders `VaultAdminDialogs` directly.

Suggested: one `<VaultAdminSurface admin target onClose />`. It is also the natural home for anything that must apply to *every* trigger — which is exactly the kind of thing that went wrong once already (the rail's Members/Roles targeting the active vault instead of the right-clicked one, fixed in #239 only after the whole-branch review caught it, because no single host owned the pairing).

## 2. The member-count message, built twice from two store snapshots

```ts
memberN > 1 ? t(…, { count: memberN - 1 }) : t(…Empty/All)
```

is written once in `VaultAdminDialogs` for the confirm message and once in `useVaultAdminActions` for the success toast, from two independently-read `membersByTeam` snapshots. Besides being the same guard twice, the two can genuinely diverge: the hook's snapshot is the pre-`fetchTeamData` one, the dialog's is re-rendered.

Suggested: derive the message once and pass it, or read the count from one place.

## Not in scope here

`counts.filter((c) => c.count > 0)` appears in both `VaultAdminDialogs` and `VaultSettingsBody`, but the downstream shapes genuinely differ (scalar sum vs. array for chips). Reviewed and deliberately left alone.

kipavy:
Fixed in #247, squash-merged to `dev` as 20a4dea4. (A PR merged to `dev` does not auto-close its issue — the closing keyword only fires on `main`.)

Issue #243, opened 2026-09-09
make-private's re-entrancy guard has no regression test


kipavy:
Follow-up from #222 / #239.

## Background

`Modal`'s keydown handler is:

```ts
if (e.key === "Enter" && onEnter) { e.stopPropagation(); onEnter(); }
```

`stopPropagation`, not `preventDefault` — so with focus on a confirm button, Enter fires `onEnter` **and** the button's native activation fires a `click`. Before #239 both closures read the same stale `busy` React state (no re-render between them in one tick), so `makePrivate` ran twice: two full copy passes writing duplicate local entities, then a second `deleteTeam` that fails and toasts an error over the success.

#239 fixed it with a `useRef` set synchronously before any `await` and cleared in `finally`. The guard is correct by construction and its placement was verified by hand.

## The gap

It shipped **without a regression test**. The bug was proven real by instrumentation, but every call-count assertion turned out vacuous in this repo's vitest/vite-node setup: an unguarded second call always dies inside a concurrent dynamic-import race (`makePrivate` opens with a long run of `await import(...)`) before reaching any mock — so `deleteTeam`, `fetchTeamData` and `saveConnection` counts look identical with or without the fix. Shipping a green test that cannot fail on the bug it names would have been worse than none, so none was shipped.

## Suggested approach

Assert on **timing**, not call counts. The guard's early `return` happens before any `await`, so with the fix the second call's promise settles almost immediately, while the unguarded version stays pending far longer (it hangs or throws deep in the import chain). Something like: fire both calls synchronously, `await Promise.resolve()` once or twice, then assert the **second** promise has already settled while the **first** has not.

That never needs the second call's chain to reach or survive the mocked services, so it sidesteps the import race entirely.

Prove it discriminates before landing it: comment out the ref guard, confirm the test fails, restore it, confirm it passes.

kipavy:
Premise changed slightly by #246 (merged to dev as e86a1359): the guard is no longer make-private's alone.

Delete became async and destructive there, so instead of giving it a second `useRef`, both verbs now share one `exclusive` helper in `useVaultAdminActions`:

```ts
const exclusive = async (run: () => Promise<void>) => {
  if (inFlight.current) return;
  inFlight.current = true;
  setBusy(true);
  try { await run(); } finally { inFlight.current = false; setBusy(false); }
};
```

[...]

Issue #242, opened 2026-09-09
The vault rail's right-click menu is wired to local vault rows only


kipavy:
Follow-up from #222 / #239.

## What is wrong

#222's premise is that the vault name in the header and a right-click on the rail open the **identical** menu. That holds for local vault rows, but `VaultSidebar` binds `onContextMenu` only inside `vaults.map(...)`, and resolves its target with `vaults.find(...)`.

The rail also renders:
- **standalone team vault rows** (`standaloneTeams.map`) — an invited member with no linked local vault
- **orphan rows** (`orphanVaultIds.map`)

Neither has the handler, so right-clicking them does nothing.

## Why it matters

`VaultHeader` already builds a `kind: "cloud"` target for `standaloneTeam` and shows the menu for it (Share / Members / Roles). So the users who most want Members and Roles — team members who do not own a local vault row — get the menu from the header but not from the rail, in the one feature whose whole point is that both triggers behave the same.

## Suggested fix

The three row wrappers are already near-identical `<div className="relative flex items-center justify-center w-full shrink-0">` blocks. Extract the wrapper, put `onContextMenu` on it once, and have the target resolution handle cloud rows the way `VaultHeader` already does (`{ kind: "cloud", vaultId: null, teamId, name }`).

That also removes a copy-paste seam rather than adding a fourth one.

kipavy:
Fixed in #247, squash-merged to `dev` as 20a4dea4. (A PR merged to `dev` does not auto-close its issue — the closing keyword only fires on `main`.)

Issue #241, opened 2026-09-09
An aborted make-private leaves partial local copies, and the toast only says nothing was deleted


kipavy:
Follow-up from #222 / #239.

## What happens

`useVaultAdminActions.makePrivate` copies every team entity back to local disk with `Promise.allSettled`, then aborts before the destructive step if any write rejected:

```ts
const rejected = writes.filter((w) => w.status === "rejected");
if (rejected.length > 0) { …copyFailedToast…; return; }
```

That guard is correct and deliberate — it exists (from #223) so a failed copy never destroys the last remaining copy of an object. It should stay.

## The gap

`allSettled` means partial success is the normal shape of a failure. If 40 of 50 writes land and 10 reject, the abort leaves **40 duplicate entities already written to local disk** with the personal `vault_id`, while the team still holds the originals.

`settings.vaults.general.makePrivate.copyFailedToast` says nothing was deleted and this is still a team vault. Both clauses are true, and both are about *deletion* — neither tells the user that partial copies now exist. Retrying duplicates them again.

## Suggested fix

Cheapest: reword the toast so it says some items were copied out and may appear twice.

Better: on abort, delete the fulfilled writes, or key the copy so a retry is idempotent.

## Also worth checking while in here

The `rejected.length` guard assumes `saveConnection` / `saveIdentity` / `saveKey` / `saveFolder` / `createSnippet` / `createPfRule` **reject** on failure rather than resolving with an error result. If any of them resolves on failure, the guard is blind and the "abort before the destructive step" property does not hold at all. This was flagged during review but not verified.

Pre-existing behaviour, moved verbatim in #239. Raised because #222 makes make-private reachable from a rail right-click.

kipavy:
Fixed in #247, squash-merged to `dev` as 20a4dea4. (A PR merged to `dev` does not auto-close its issue — the closing keyword only fires on `main`.)

PR #245, opened 2026-09-09
fix(vaults): tell the truth about make-private, and stop it losing data (#68)

kipavy:
The consent modal for turning a private vault into a team vault ended with "There's no way back to private." That is false: `handleMakePrivate` in `VaultsSection.tsx` converts a team vault back, and its gate (`isTeam && isOwner && detail.kind === "local"`) describes exactly the person reading the modal. The line now says what is actually true — you can go back, but it removes everyone's access and does not take back the copies they already hold.

Making that promise honest meant looking at the handler behind it, which had three ways to fail quietly:

- Every entity was re-written to local disk under one `Promise.allSettled` whose rejections were discarded, and the team was deleted afterwards. Any write that failed had its last remaining copy deleted off the server a moment later. A failed write now aborts before anything destructive runs.
- `deleteTeam(teamId).catch(() => {})` meant a failed server call still unlinked the vault locally and reported "Vault is now private" while every member kept their access. The delete now runs before the local teardown, and a failure stops the whole operation — the server call is the part that actually revokes access.
- The outer catch only reached the console, so an unexpected failure showed the user nothing at all. It now raises a toast.

Error reasons go through `userFacingReason`, extracted from `vaultShare`'s `inviteFailureReason` (which now delegates to it): a raw transport failure carries the server URL in its message, and that must not reach the UI.

## Tests

`src/components/settings/sections/VaultsSection.MakePrivate.test.tsx` — 5 tests, covering the abort-before-destructive path, the `deleteTeam` failure path, and the toast on unexpected failure. `tsc --noEmit` clean.

## Known follow-ups

Filed rather than folded in, so this PR stays one change:

- #240 — vault delete says "permanently deletes… cannot be undone", but `removeVault` leaves the objects on disk
- #241 — an aborted make-private leaves partial local copies, and the toast only says nothing was deleted
- #242 — the vault rail's right-click menu is wired to local vault rows only
- #243 — the re-entrancy guard has no regression test
- #244 — two duplicated shapes left by the rework

https://claude.ai/code/session_01P77tdVQCRa1zkDCCpqomgr


kipavy:
Duplicate — this work already landed on dev as #223 (a3abe7a5). The local branch read as unmerged because #223 was squash-merged; the diff against a3abe7a5 is empty. Follow-ups #240-244 remain open.

PR #251, opened 2026-09-09
fix(vaults): stop make-private reporting itself as a removal (#249)

### LINKED ISSUE #249: Make-private tells the owner they "were removed" and runs the offboarding secret wipe against their own objects
Found while live-verifying #247 (#241/#242/#243/#244). Pre-existing — reproduced on both `dev@20a4dea4` and the pre-#247 build, so #247 neither caused nor fixed it.

## What happens

Convert a private vault to a team vault, then use **Make private…** on it as the owner. The operation succeeds — the team is deleted, the objects come back to local disk — and then the notification inbox raises:

> You were removed from **E2E Vault**. Its hosts, keys and snippets are gone from this device, and credentials you used there should be rotated.

(`notifications.inbox.membershipEnded.message`, `src/i18n/locales/en/notifications.json:79`.)

Every clause is false for this user. They were not removed, they *own* those objects, the hosts and keys are on this device (that is the whole point of make-private), and there is nothing to rotate.

## Where it comes from

`makePrivate` (`src/components/vault-admin/useVaultAdminActions.ts`) calls `deleteTeam(teamId)` directly. The next sync's team-list diff sees the team gone and treats it as an offboarding:

- `src/services/sync.ts:947` — `onTeamRemoved: (tid) => offboardFromTeam(tid, …)`
- `src/services/sync.ts:888` — the notify step fires `notifyMembershipEnded(teamName)` unless `departedVoluntarily(tid)`
- `src/services/teamOffboarding.ts:74` — `voluntaryDepartures.add(teamId)` is only reached from `departMembers`

`makePrivate` never goes through `departMembers`, so it never marks intent, so its own deletion reads as a kick.

## The more serious half

`offboardFromTeam` does not just notify. Before anything else it runs `clearTeamStoresAndSecrets(tid)` (`src/services/sync.ts:858`), which derives the keychain keys to wipe **from the object ids still held in the team store slices** (`src/services/teamVaultSync.ts:566`):

```ts
...conns.flatMap((c) => [`key:${c.id}`, `password:${c.id}`, `passphrase:${c.id}`]),
...keys.flatMap((k) => [`key:${k.id}:private`, `key:${k.id}:public`, `key:${k.id}:passphrase`]),
...identities.map((i) => `identity:${i.id}:password`),
```

Since #247, the local copies live under **exactly those ids** — so that key list is now a list of the user's own live credentials. The only reason they survive is ordering: `makePrivate` clears the team slices in the same tick `deleteTeam` resolves, so by the time the wipe runs it derives an empty list and deletes nothing.

That held in the live run — all three seeded secrets (`password:<id>`, `key:<id>:private`, `identity:<id>:password`) read back fine afterwards. But it is correctness by ordering between an `await` continuation and a sync poll, not correctness by construction, and it is the kind of thing that is fine until someone adds an `await` between `deleteTeam` and the clears.

## Suggested fix

Make `makePrivate` declare intent, so its own deletion is not offboarding at all. `voluntaryDepartures` already exists for precisely this shape of race — either export a marker from `teamOffboarding` and call it before `deleteTeam`, or give `offboardFromTeam` a way to recognise a team this client deleted as owner.

That kills the wrong notice and removes the dependency on clear-before-wipe ordering in one change. A regression test should assert that a make-private raises no `membershipEnded` notice and that the secrets are still readable after the offboarding path has run.

## Reproduction

Isolated `voltius-e2e-*` stack, one cloud account, seven object kinds seeded in a vault with three real keychain secrets, converted to a team vault, then Make private… from the rail's right-click menu. The notice appears within one sync interval.

kipavy:
Fixes #249.

## The wrong notice

Make-private deletes the team, the server echoes it back as a `membership_changed`, and the team-list diff read this client's own deletion as a kick — so the owner got `notifications.inbox.membershipEnded.message`: *"You were removed from **E2E Vault**. Its hosts, keys and snippets are gone from this device, and credentials you used there should be rotated."* Every clause is false for the user who just took those objects private.

## The wipe

`offboardFromTeam` runs `clearTeamStoresAndSecrets` before anything else, and that derives its keychain keys from the object ids still in the team store slices. Since #247 the local copies live under exactly those ids, so that key list was a list of the user's own live credentials. They survived only because `makePrivate` cleared the slices in the same tick `deleteTeam` resolved — correctness by ordering between an `await` continuation and a sync poll.

## The change

- `voluntaryDepartures` becomes `selfDepartures`, a map of `"leave" | "self-deleted"`. A `leave` behaves exactly as before: full offboarding, notice suppressed. A `self-deleted` team is not an offboarding at all — no wipe, no `forbidden` status, no notice.
- `makePrivate` marks `self-deleted` **before** `deleteTeam`, and now drops the team's own store slices (`removeTeam`) itself rather than leaving that to the sync poll — read after the toast has read the roster, never before.
- `vaultConvert`'s rollback delete is marked the same way, for the same reason: the objects it uploaded still exist locally under those ids.
- Defence in depth: `clearTeamStoresAndSecrets` never wipes a keychain entry a *local* object of the same id still owns. A genuine removal never reaches that filter — a member losing access holds no local copy of the team's objects.

## Tests

Each new test was verified by reverting its fix and watching it fail:

- `sync.teamRemovedFailures` — a self-deleted team is not offboarded, with an unmarked control team in the same event proving the harness would have caught a wipe; and a `leave` still wipes with only the notice suppressed.
- `teamSecretWipe` — never wipes a secret a local object of the same id owns.
- `useVaultAdminActions` — the delete is marked before it is sent; the team's store slices go with it, after the roster is read.
- `vaultConvert` — the rollback delete is marked too.

`services` 1175, `stores` 459, `components` 910, `tsc --noEmit` exit 0.

### DIFF
--- a/src/components/vault-admin/useVaultAdminActions.test.tsx
+++ b/src/components/vault-admin/useVaultAdminActions.test.tsx
@@ -3,6 +3,7 @@ import { render, screen, cleanup, fireEvent, waitFor } from "@testing-library/re
 
 const h = vi.hoisted(() => ({
   deleteTeam: vi.fn(async (_id: string) => {}),
+  markSelfDeparture: vi.fn(),
   addToast: vi.fn(),
   fetchTeamData: vi.fn(async (_id: string) => {}),
   clearTeamKeyCache: vi.fn(),
@@ -34,6 +35,7 @@ vi.mock("@/services/teamService", () => ({
   listPendingInvitations: vi.fn(async () => []),
   revokePendingInvitation: vi.fn(),
 }));
+vi.mock("@/services/teamOffboarding", () => ({ markSelfDeparture: h.markSelfDeparture }));
 vi.mock("@/hooks/useVaultContents", () => ({ useVaultContents: () => [] }));
 vi.mock("@/hooks/useUIContributions", () => ({ useUIContributions: () => [] }));
 vi.mock("@/components/shared/ContentCounts", () => ({ ContentCounts: () => null }));
@@ -198,6 +200,33 @@ test("the happy path deletes the team, unlinks the vault and reports success", a
   expect(h.clearTeamConnections).toHaveBeenCalledWith("t1");
   expect(h.clearTeamKeyCache).toHaveBeenCalled();
   expect(messages()).toEqual(["settings.vaults.general.madePrivateToastEmpty"]);
+});
+
+test("the team's own store slices go with it", async () => {
+  // The offboarding path used to drop these, and it no longer runs for a team
+  // its owner deleted (#249); the roster the toast reports on is read first.
+  useTeamStore.setState({
+    membersByTeam: { t1: [{ user_id: "me" } as never, { user_id: "ana" } as never] },
+  });
+
+  clickMakePrivate();
+
+  await waitFor(() => expect(onDone).toHaveBeenCalled());
+  expect(useTeamStore.getState().teams).toEqual([]);
+  expect(useTeamStore.getState().membersByTeam.t1).toBeUndefined();
+  expect(messages()).toEqual(["settings.vaults.general.madePrivateToast"]);
+});
+
+test("the delete is marked as this client's own before it is sent", async () => {
+  // Unmarked, the server's echo of this delete reads as a kick: the wrong
+  // notice, and an offboarding wipe aimed at the copies just adopted under the
+  // same ids. Marking after the round trip would be too late (#249).
+  clickMakePrivate();
+
+  await waitFor(() => expect(h.deleteTeam).toHaveBeenCalled());
+  expect(h.markSelfDeparture).toHaveBeenCalledWith("t1", "self-deleted");
+  expect(h.markSelfDeparture.mock.invocationCallOrder[0])
+    .toBeLessThan(h.deleteTeam.mock.invocationCallOrder[0]);
 });
 
 test("an unexpected failure surfaces as an error toast, not just a console line", async () => {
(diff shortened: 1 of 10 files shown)


kipavy:
**Live-verified** on an isolated `voltius-e2e-249-*` stack (own postgres + `voltius-server:latest`, self-hosted mode) with my own `tauri-headless-249` container mounting this worktree, one real cloud account, driven over raw WebDriver. Prod untouched throughout.

Seeded a vault with a connection, an SSH key and an identity, plus three real keychain secrets (`password:<connId>`, `key:<keyId>:private`, `identity:<identId>:password`), converted it to a team vault (3/3 rows in `team_vault_objects` and 3 in `team_vault_secrets`, all under the original ids), then ran **Make private…** from the rail's right-click menu.

**With the fix — two full convert → make-private cycles:**
- notification inbox **empty** (checked 45 s and 60 s after, well past the `membership_changed` echo);
- all three secrets still readable, unchanged values;
- objects on disk: exactly one of each, original ids, back under the private vault, `notes` preserved;
- `teams` and `membersByTeam` cleared, vault status not `forbidden`.

**Control — same box, same flow, sources swapped back to `dev@20a4dea4`** (`git checkout 20a4dea4 -- <the five files>`, reload, repeat): the bug reproduces within ~4 s of the confirm —

```
inbox:  ["membership-ended:E2E249"]
toast:  "You were removed from E2E249. Its hosts, keys and snippets are gone from this de…"
```

so the `membership_changed` echo really does arrive on this harness, and the clean run above is the fix, not a missing event.

Bonus: a conversion that failed mid-run (`initTeamVaultKey` → "Vault is locked" right after a reload) exercised the `vaultConvert` rollback path on the fixed build — its `deleteTeam` raised no notice either.

PR #253, opened 2026-09-10
fix: stop macOS capitalising the first letter of SSH usernames

### LINKED ISSUE #252: [BUG] Host username input field auto capitalize on first letter
## Steps to Reproduce
1. Go to 'Personal vault'
2. Click on 'New Host'
3. Scroll down to 'Identity'
4. Input a username `abcd`
5. Click outside, the input become `Abcd`

## Expected Behavior
The username input should stays `abcd` without first letter capitalization

## Screenshots / Video

https://github.com/user-attachments/assets/597c87f2-dc86-471b-9479-b7f531ffc5ac

## Environment Info
*Please tell us what device you are using so we can reproduce the bug.*

* **OS:** macOS 26.6.2
* **App Version:** v0.32.1

**Additional context**

snowfluke:
Closes #252.

## What

macOS capitalises the first letter of a plain text input when it loses focus, so typing `abcd` into a host's Identity username and clicking away saved `Abcd`. An SSH username is case-sensitive, so the host was stored with a different login than the one that was typed, and the connection then failed for a reason nothing on screen explained.

The keychain identity form has the same field with the same defect, and a keychain identity feeds the same SSH login, so it is fixed here too.

## Which attribute actually fixes it

Worth reading before reviewing, because the obvious attribute is the one that does nothing.

`autocapitalize` is the attribute whose name matches the bug, and on its own it does not fix it. I measured all three in a bare `WKWebView`, the engine Tauri renders through, with the macOS "Capitalize words automatically" setting confirmed on (`NSAutomaticCapitalizationEnabled = 1`). Typed `abcd` into each field, clicked away, read the value back:

| Attribute on the input | Value after blur | |
| --- | --- | --- |
| none (control) | `Abcd` | capitalised |
| `autocapitalize="off"` | `Abcd` | **capitalised** |
| `autocorrect="off"` | `abcd` | left alone |
| `spellcheck="false"` | `abcd` | left alone |
| all three | `abcd` | left alone |

So macOS routes this through text checking, not through autocapitalize. That matches the MDN note that `autocapitalize` has no effect when typing on a hardware keyboard.

Two further things that fell out of measuring it:

- **Safari does not reproduce this at all.** The control stays `abcd` there even with the OS setting on, so Safari must be opting web content out of the text-checking pipeline. Anyone verifying this needs a WKWebView, not a browser tab.
- **`autocapitalize="off"` still earns its place**, just not on desktop. It is the attribute that governs virtual keyboards, which is what the Android build gets typed into.

Hence all three ship together rather than only the one that fixes macOS.

## How

`src/components/shared/Panel.tsx` gains a `formIdentifierProps` constant next to the existing `formInputClass` / `formInputStyle` exports:

```ts
export const formIdentifierProps = {
  autoCapitalize: "off",
  autoCorrect: "off",
  spellCheck: false,
} as const;
```

It is spread onto the username input of `ConnectionForm` and of `IdentityForm`. `UsernamePromptPanel` already carried these same three attributes written out inline, so it now uses the shared constant rather than leaving a third copy of the same shape, per the DRY rule in `CLAUDE.md`. The measured table above is recorded as a comment on the constant, so the next person does not have to re-derive why an apparently redundant attribute is there.

## Scope

Limited to usernames, because that is the only field where a capital letter changes what the app does rather than how it looks:

| Field | Capitalised | Effect |
| --- | --- | --- |
| Username | `abcd` to `Abcd` | Different SSH login. Fixed here. |
| Hostname | `srv` to `Srv` | Cosmetic. DNS is case-insensitive. Left alone. |
| Connection or snippet name | `prod` to `Prod` | Wanted. Left alone. |

Happy to widen it to hostnames in this PR if you would rather have them consistent.

## Testing

Please lean on CI. I do not have the toolchain set up locally, so **none of the required checks were run**: no `cargo fmt`, `cargo clippy`, `cargo build` or `pnpm run build`, and the two tests below have not been executed.

Added to the existing characterisation suites:

- `connectionForms.characterisation.test.tsx` asserts the SSH username input renders `autocapitalize="off"`, `autocorrect="off"` and `spellcheck="false"`.
- `keychainForms.characterisation.test.tsx` asserts the same for the identity username input.

All three are asserted deliberately. Since `autocorrect` and `spellcheck` are the two carrying the fix, a test that checked only `autocapitalize` would pass while the bug came straight back.

The change is attributes-only, with no behavioural code touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- The username of a host and of a keychain identity no longer picks up a
+  capital first letter on macOS. Typing `abcd` and clicking away left `Abcd`
+  behind, and an SSH username is case-sensitive, so the host was saved with a
+  different login than the one that was typed. Fields you name yourself, such
+  as a connection or a snippet name, keep the normal macOS behaviour.
+
 ## [0.32.1] - 2026-09-07
 
 ### Fixed
--- a/src/components/connections/ConnectionForm.tsx
+++ b/src/components/connections/ConnectionForm.tsx
@@ -41,6 +41,7 @@ import {
   formInputStyle,
   formLabelClass,
   formLabelStyle,
+  formIdentifierProps,
 } from "@/components/shared/Panel";
 import { SecretInput, TagsAndFolderFields } from "@/components/shared/vaultObjectForm";
 import {
@@ -580,6 +581,7 @@ const ConnectionForm = forwardRef<ConnectionFormHandle, Props>(function Connecti
                     value={username}
                     onChange={(e) => { markDirty(); setUsername(e.target.value); }}
                     placeholder="root"
+                    {...formIdentifierProps}
                   />
                 </div>
 
--- a/src/components/connections/connectionForms.characterisation.test.tsx
+++ b/src/components/connections/connectionForms.characterisation.test.tsx
@@ -373,3 +373,15 @@ test("a dirty edit marks the form dirty on both forms", () => {
   fireEvent.click(document.querySelector("[data-tag-selector]")!);
   expect(serial.ref.current!.isDirty()).toBe(true);
 });
+
+// #252: macOS capitalised the first letter of the SSH username on blur, so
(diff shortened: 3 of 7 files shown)


snowfluke:
Proof:

# Safari

image

# Tauri WKWebView

image

Issue #16, opened 2026-06-15
Feature: Integrated SFTP File Editor using CodeMirror 6


kipavy:
## Description

To provide a seamless SSH workflow, we need an integrated file editor within the application. This will allow users to open, view, edit, and save remote files directly inside the client without relying on terminal editors like nano or vim.

We will use @uiw/react-codemirror (CodeMirror 6) in the React frontend due to its modular architecture, light footprint, and robust extension ecosystem.

## User Stories

- As a user, I want to click on a remote file in the file explorer and have it open in an integrated editor tab.
- As a user, I want syntax highlighting automatically applied based on the file extension.
- As a user, I want to press `Ctrl + S` (or `Cmd + S`) to instantly save changes back to the remote server.
* **Auto-Save Toggle:** Provide a preference setting (Global and Editor-specific toolbar checkmark) to switch between manual `Ctrl + S` updates and a 1000ms debounced auto-save.
* **Text History Isolation:** Maintain CodeMirror's native text history instance per document tab. Text mutations will use native `history()` bindings, preventing keystrokes from saturating the global application state machine.

simonlpaige:
Good points on the approach. A few technical considerations from examining the codebase:

1. **File size handling**: The current SFTP implementation in `src-tauri/src/sftp/stream.rs` uses chunked reads, but the frontend doesn't currently have a size threshold for the editor. Files >10MB will likely freeze the UI if loaded entirely into CodeMirror. Consider adding a configurable size limit (default ~5MB) with a warning before opening, or implement streaming/lazy loading for large files.

2. **Binary file detection**: The file explorer (`src/components/explorer/FileList.tsx`) currently shows all files as openable. You'll want to check the file's MIME type or first few bytes before loading into CodeMirror – opening a `.png` or `.o` file will corrupt the editor state. The `ssh2` crate's `stat()` can give you file metadata before fetching content.

3. **Save conflict handling**: With auto-save on a 1s debounce, if the user switches tabs while a save is in-flight, you'll need to queue or cancel pending writes. The current Tauri command `write_file` in `src-tauri/src/sftp/commands.rs` doesn't handle concurrent writes to the same path – you'd want a per-file mutex or a write queue to prevent race conditions.

[...]

Issue #17, opened 2026-06-22
unable to login in old Cisco switches 


VTX-Reflectiz:
Description:
I need to connect via SSH to legacy network equipment (Cisco switches with old IOS firmware) 
that only support deprecated, weak algorithms. Voltius currently has no way to enable these, 
either via GUI or command-line flags, making it impossible to connect to such devices.

Error received:
Connection failed: No common Kex algorithm - ours: ["mlkem768x25519-sha256", 
"curve25519-sha256", "curve25519-sha256@libssh.org", "diffie-hellman-group-exchange-sha256", 
"diffie-hellman-group18-sha512", "diffie-hellman-group17-sha512", "diffie-hellman-group16-sha512", 
"diffie-hellman-group15-sha512", "diffie-hellman-group14-sha256"], theirs: ["diffie-hellman-group1-sha1"]

I had the subsequent error using CMD and with the help of Claude I managed to edit the .ssh file in the windows user directory.
Subsequent errors after working around Kex (via OpenSSH) on the same device:
- No matching host key type found. Their offer: ssh-rsa
- No matching cipher found. Their offer: 3des-cbc
- No matching MAC found. Their offer: hmac-sha1, hmac-sha1-96, hmac-md5, hmac-md5-96

Request:
Could you add an option (ideally per-session, to avoid weakening security globally) to enable 
legacy algorithms when needed? Something equivalent to OpenSSH's config:

Host *
    KexAlgorithms +diffie-hellman-group1-sha1
    HostKeyAlgorithms +ssh-rsa
    Ciphers +3des-cbc
    MACs +hmac-sha1,hmac-md5

This would help users who manage older infrastructure (e.g. old Cisco switches) where firmware 
upgrades aren't always immediately feasible.

Environment:
- Voltius version: [0.5.1]
- OS: Windows 11
- Target device: Cisco Catalyst switch (legacy IOS)

kipavy:
Hi, thank you for your report, I'll see if I can add back support for older algorithms

kipavy:
[...]

PR #254, opened 2026-09-10
feat(teams): rotate the team DEK on membership change — client (#217)

kipavy:
## Summary

Client half of #217, depends on the already-merged-and-deployed server PR (VoltiusApp/server#16). Makes the client read/write team-vault ciphertext version-aware, and adds the opportunistic rotate-and-drain pass that runs on team membership changes.

- Version-aware key cache: `getCachedTeamKeyVersion`, `getTeamVaultKeyAtVersion`, on top of the unchanged `getTeamVaultKey` signature.
- Version-aware encode/decode for objects, secrets, and the legacy vault blob.
- `teamKeyRotation.ts`: `checkAndRotateTeamKey` — on every `team_members:<teamId>` event, either mints a new epoch (wraps a fresh 32-byte DEK for every current member with a public key) or resumes draining a previous rotation (re-encrypts stale objects, secrets, and the legacy blob under the current epoch), never both in the same call.
- Wired into `sync.ts`'s realtime handler, fire-and-forget.

Design: `docs/superpowers/specs/2026-09-09-team-dek-rotation-design.md`. Plan: `docs/superpowers/plans/2026-09-09-team-dek-rotation-client.md`.

**Two rounds of review caught real bugs before this reached anyone**, both worth knowing about since they're the kind unit tests structurally struggle to catch:
- Task-level: the rotation reference code re-wrapped the *existing* DEK instead of generating a fresh one (zero cryptographic effect), and the object-reencrypt loop double-encrypted ciphertext instead of decrypt-then-re-encrypt (would have destroyed metadata on every rotated row). Both fixed and re-verified.
- Final whole-branch review: a cold-cache ordering bug in `decodeObjectMetadata` (stale rows silently dropped on app restart during a drain), and the legacy vault blob was never wired into the drain at all — a team with one would rotate once and then be permanently stuck unable to rotate again. Both fixed and re-verified.

## Test plan

- [x] 89/89 passing across all 13 directly-relevant test files (`teamVaultSync*`, `teamKeyRotation`, `teamObjectEnvelope`, `teamVaultSecrets`, `logger`)
- [x] `tsc --noEmit` clean
- [x] Full repo-wide `vitest run` — green on CI (`build` job, "Frontend tests" step) for `ca824d69`, [run 34444158144](https://github.com/VoltiusApp/voltius/actions/runs/34444158144). The same job's `cargo fmt --check`, `clippy -D warnings` and `cargo test --workspace` are green too. Local runs were OOM-killed on this box under system-wide memory pressure, unrelated to this change; CI is the confirmation.
- [ ] Live two-account e2e verification against a real server (rotate, confirm both accounts stay in sync through the transition) — not yet done, flagged as the natural next step before this ships to users.

### DIFF
--- a/src/i18n/locales/en/common.json
+++ b/src/i18n/locales/en/common.json
@@ -114,6 +114,8 @@
       "failedToDeleteTeam": "Failed to delete team: {{status}}",
       "failedToListMembers": "Failed to list members: {{status}}",
       "failedToListVaultKeyHolders": "Failed to list vault key holders: {{status}}",
+      "failedToCheckRotationStatus": "Failed to check rotation status: {{status}}",
+      "failedToRotateVaultKey": "Failed to rotate vault key: {{status}}",
       "failedToAddMember": "Failed to add member: {{status}}",
       "failedToRemoveMember": "Failed to remove member: {{status}}",
       "failedToAssignRole": "Failed to assign role: {{status}}",
--- a/src/i18n/locales/fr/common.json
+++ b/src/i18n/locales/fr/common.json
@@ -114,6 +114,8 @@
       "failedToDeleteTeam": "Échec de la suppression de l'équipe : {{status}}",
       "failedToListMembers": "Échec du chargement des membres : {{status}}",
       "failedToListVaultKeyHolders": "Échec du chargement des détenteurs de clé de coffre : {{status}}",
+      "failedToCheckRotationStatus": "Échec de la vérification de l'état de rotation : {{status}}",
+      "failedToRotateVaultKey": "Échec de la rotation de la clé de coffre : {{status}}",
       "failedToAddMember": "Échec de l'ajout du membre : {{status}}",
       "failedToRemoveMember": "Échec de la suppression du membre : {{status}}",
       "failedToAssignRole": "Échec de l'attribution du rôle : {{status}}",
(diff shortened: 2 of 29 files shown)


kipavy:
Addressed the 8 review findings that were client-side bugs (commit bae1cd1d):

- **Rotation never retried on reconnect** — `onTeamLogin` now also calls `checkAndRotateTeamKey`, not just the realtime `team_members` handler.
- **Wrong-epoch fetch for legacy secrets** — the stale-secrets reencrypt mapper and `hydrateTeamVaultSecrets` now apply the epoch-1 fallback before fetching, instead of requesting `vault-key/undefined`.
- **Historical epoch lookup broke after member removal** — the wrapping user is now resolved directly via `getUserPublicKey`, not the current member roster (which would never find someone since removed — the very event that triggers rotation).
- **Version cache not set alongside key cache** — `initTeamVaultKey`'s fresh-key branch now caches epoch 1, keeping the two caches in lockstep as documented.
- **Uncaught batch-PUT failure aborted the whole drain** — `settleInBatches`'s `onBatch` call is now itself caught, so one bad batch can't block the rest.
- **Deleted-object secrets still reencrypted** — `objectTypeById` now excludes soft-deleted objects, matching `staleObjects`'s own filter.
- **Legacy-blob drain serialized unnecessarily** — it now runs concurrently with the object/secret list fetches via `Promise.all`.
- **Triplicated epoch-fetch logic already drifted** — the `teamVaultSecrets.ts` copy now applies the same `?? 1` normalization as the other two call sites.

Not changed: the debounce/cooldown suggestion for `checkAndRotateTeamKey` — reverted after prototyping it, since it broke several existing dedup tests for a low-value efficiency gain and isn't worth the complexity.

Separately: **the actual reason removal never rotated the DEK live was a server-side gap**, not a client bug — `get_rotation_status`'s `stale` check had no way to detect a removal at all. Fixed and live-verified in VoltiusApp/server#17.

All 339 tests in `src/services/team*`/`sync*` pass, `tsc --noEmit` clean.

PR #285, opened 2026-09-15
feat(i18n): add Turkish localization

alkank:


kipavy:
Thanks, it'll include this in next release, coming today

